PR #49, opened 2024-11-15
feat: init openshift reference arch

leiicamundi:
Related to https://github.com/camunda/team-infrastructure-experience/issues/509

This PR adds the OpenShift mono-region reference arch, that is also used in the documentation.
Integration is still in progress so references may be broken

8.7 routes are currently broken due to https://camunda.slack.com/archives/CSQ2E3BT4/p1733781541218609

Tests will come later in another PR


Langleu (approved):
Ran through the files with the guides, and it worked well for the Terraform part.
The helm chart should probably work, but my snapshot release was broken, so it never became healthy.
I used the routes approach, which was working; for example, Identity was exposed, but it was not working past the login screen to due related application problems.

Overall, the guide is working to the extent it could. The huge value files and additions are a bit of a hard pill to swallow. It takes a bit of patience to get through / understand.
This isn't an issue and don't have an alternative.

PR #52, opened 2024-11-21
chore: split up ec2 tests from env creation

Langleu:
- splits up creation / teardown from the running all tests
- allows us to add e2e tests in different languages by QA by reusing the same environment.


Langleu:
@leiicamundi some small changes to ease the weekly renovate PRs + split up creation and teardown, so we / QA can nicely integrate the e2e tests on top while reusing the same environment.

PR #77, opened 2025-01-03
feat: infraex-465 configure new camunda exporter

Langleu:
related to https://github.com/camunda/team-infrastructure-experience/issues/465

configured the new camunda exporter.
Left out the cluster_name for now to see whether I really need to configure it or not but added it already to the output if required.


Langleu:
closing as I missed the SNAPSHOT suffix.

PR #87, opened 2025-01-14
lint(pre-commit): add trivy check on terraform directory (issues ignored)

bkenez:
Source issue: https://github.com/camunda/team-infrastructure-experience/issues/37

Extending the scope of the above issue to this repository as well, implementing a trivy scan of terraform files within was directory, excluding fixtures (formerly tfsec).

Issues are ignored for now and listed in .trivyignore

### DIFF
--- a/.lint/trivy/trivy-scan.sh
+++ b/.lint/trivy/trivy-scan.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -euxo pipefail
+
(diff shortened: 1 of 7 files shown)


bkenez:
It's the security group rules that were missing them, not the group themselves, but added it 👍

bkenez:
Adding the descriptions breaks the golden file check it seems.

PR #89, opened 2025-01-23
feat(aws): openshift dual region hcp

leiicamundi:
This PR introduces the necessary IaC files to setup OpenShift in a dual region with the associated instructions to install Camunda 8.6 with Submariner.

As the rescope has been announced, the 8.7 instructions are using the same 8.6 parameters.

A separate issue will be created for the 8.8 parameters once this one is validated

Regarding the tests, the `MultiClusterHub` suffers from an interesting bug, when it is created manually using the UI it works, but using manifest it fails.
I'll post the tests later in a separate issue.

### DIFF
--- a/aws/rosa-hcp-dual-region/procedure/camunda/8.7/generate_zeebe_helm_values.sh
+++ b/aws/rosa-hcp-dual-region/procedure/camunda/8.7/generate_zeebe_helm_values.sh
@@ -66,7 +66,7 @@ if [[ "$mode" == "failover" ]]; then
   read -r -p "Enter the region that was lost, values can either be 0 or 1: " lost_region
   if [[ "$lost_region" != "0" && "$lost_region" != "1" ]]; then
     echo "ERROR: Invalid region $lost_region provided for the lost region. Please provide either 0 or 1 as input value."
-    return 1
+    exit 1
   fi
 fi
 
@@ -79,15 +79,15 @@ cluster_size=$((ZEEBE_CLUSTER_SIZE))
 
 if (( cluster_size % 2 != 0 )); then
   echo "ERROR: Cluster size $cluster_size is an odd number and not supported in a multi-region setup (must be an even number)"
-  return 1
+  exit 1
 fi
 if (( cluster_size < 4 )); then
   echo "ERROR: Cluster size $cluster_size is too small and should be at least 4. A multi-region setup is not recommended for a small cluster size."
-  return 1
+  exit 1
 fi
 if [[ "$namespace_0" == "$namespace_1" ]]; then
   echo "ERROR: Kubernetes namespaces for Camunda installations must be called differently"
-  return 1
+  exit 1
 fi
 
 # Generate values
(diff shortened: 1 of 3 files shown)


aws/rosa-hcp-dual-region/camunda-version/8.7/s3-elastic-backup/backup-bucket.tf
```diff
@@ -0,0 +1,135 @@
...
+}
+
+resource "aws_s3_bucket" "log_bucket" {
+  bucket = "${var.bucket_name}-log"
+}
+
+resource "aws_s3_bucket_acl" "log_bucket" {
+  acl    = "log-delivery-write"
+  bucket = aws_s3_bucket.log_bucket.id
+}
```
Langleu, line 71: as mentioned in the guide, the acl was causing problems.
I removed it in the end to continue

aws/rosa-hcp-dual-region/camunda-version/8.7/region2/cluster_region_2.tf
Langleu, line 1: as mentioned in the docs, since it's in a named subfolder we could just keep the file name generic

aws/rosa-hcp-dual-region/camunda-version/8.7/procedure/provision/gather-peering-info.sh
Langleu, line 1: not the biggest fan of this approach from a ux perspective

aws/rosa-hcp-dual-region/camunda-version/8.7/procedure/camunda/generate_zeebe_helm_values.sh
Langleu, line 1: has potential to define the total amount of zeebe brokers via environmen variable as well.
Would be easier if one just follow along to the guide as that topic was never discussed and it's just assumed the defaults are used.

aws/rosa-hcp-dual-region/camunda-version/8.7/procedure/camunda/export_environment_prerequisites.sh
```diff
@@ -0,0 +1,33 @@
...
+# The names of your OpenShift clusters in regions 1 and 2
+export CLUSTER_1_NAME=""
+export CLUSTER_2_NAME=""
+
+# The OpenShift namespaces for each region where Camunda 8 should be running
+# Namespace names must be unique to route the traffic
+export CAMUNDA_NAMESPACE_1="camunda-$CLUSTER_1_NAME"
+export CAMUNDA_NAMESPACE_2="camunda-$CLUSTER_2_NAME"
+
+# The backup bucket access variables
```
Langleu, line 24: as mentioned could link to the related doc part of the aws openshift dual-region setup guide.
Just as an optional assistance for people that follow end2end

aws/rosa-hcp-dual-region/camunda-version/8.7/region1/config.tf
Langleu, line 1: as mentioned in the docs, maybe there could be a visual differentiation like utilizing a comment to say region 1 / region 2.
So if using the tabs it's more visible that it's not the same file, even though it kinda is.
It's visually challenging with the tabs where it starts and stops in the docs.

Langleu (changes requested):
I didn't go in detail through every little thing as I followed the docs along and left some notes in the docs related to these files already.

Langleu (approved):

PR #106, opened 2025-02-24
feat: add README.md

leiicamundi:
We don't have one 😱 

Preview available at https://github.com/camunda/camunda-deployment-references/tree/feature/readme


Langleu (approved):
👍 

was adding in the merge project also one - https://github.com/camunda/camunda-deployment-references/tree/merge-branch .
But let's just merge this and I can readjust it later.

PR #125, opened 2025-03-06
feat: add Azure AKS reference

bkenez:
DRAFT

Still a work in progress, just created to ensure compliance with pre-commit hooks and to have something up to start drafting on


azure/aks/main.tf
Langleu, line 1: I'd suggest having a look at [the readme](https://github.com/camunda/camunda-deployment-references?tab=readme-ov-file#naming-convention) in regards to folder structure:

Langleu (commented):
suggestion to rebase or merge based on main, it's lagging a bit behind.

azure/aks/main.tf
```diff
@@ -0,0 +1,70 @@
...
+# Local resource to indicate test readiness
+resource "local_file" "deployment_complete" {
+  content  = "Deployment completed on ${timestamp()}"
+  filename = "${path.module}/deployment_complete.txt"
+
+  depends_on = [
+    module.aks,
+    module.postgres_db
+  ]
+}
```
Langleu, line 70: not sure what this is about and what for.
Likely not end up in the final, if required for testing could be achieved with a fixture / override.

azure/aks/modules/network/main.tf
```diff
@@ -0,0 +1,83 @@
...
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = azurerm_virtual_network.aks_vnet.name
+  address_prefixes     = ["10.1.0.0/24"]
+}
+
+resource "azurerm_subnet" "db_subnet" {
+  name                 = "${var.resource_prefix}-db-subnet"
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = azurerm_virtual_network.aks_vnet.name
+  address_prefixes     = ["10.1.1.0/24"]
```
Langleu, line 20: address_space / address_prefixes should probably be configurable.

azure/aks/modules/network/main.tf
```diff
@@ -0,0 +1,83 @@
+resource "azurerm_virtual_network" "aks_vnet" {
+  name                = "${var.resource_prefix}-vnet"
+  location            = var.location
+  resource_group_name = var.resource_group_name
+  address_space       = ["10.1.0.0/16"]
```
Langleu, line 5: do I understand it correctly that the vnet address_space is for the whole network and must include whatever you assign to your subnets?
Besides that it's dynamically used for assigning pod IPs etc.

While the `aks_subnet` is solely for VM purposes? So I could run a max of 256 nodes?
Same for the DBs I assume

Langleu: I see the cluster having a different range.

```
Pod CIDR
10.244.0.0/16
Service CIDR
10.0.0.0/16
```

How does it relate to the vnet?

Probably an architecture diagram could help to get an overview.

azure/aks/modules/network/main.tf
```diff
@@ -0,0 +1,83 @@
...
+    destination_address_prefix = "*"
+  }
+}
+
+resource "azurerm_subnet_network_security_group_association" "aks_nsg_association" {
+  subnet_id                 = azurerm_subnet.aks_subnet.id
+  network_security_group_id = azurerm_network_security_group.aks_nsg.id
+}
+
+# Simple DNS zone for testing
```
Langleu, line 70: wondering what the plan is for all the `for testing` adjustments?

azure/aks/modules/aks/providers.tf
Langleu, line 1: small suggestion to align the file names with the other sources.

azure/aks/modules/aks/main.tf
```diff
@@ -0,0 +1,62 @@
...
+  dns_prefix          = "${var.aks_cluster_name}-dns"
+  kubernetes_version  = var.kubernetes_version
+  tags                = var.tags
+
+  # Identity management
+  identity {
+    type = "SystemAssigned"
+  }
+
+  # System node pool configuration
```
Langleu, line 14: just a suggestion to add a link to https://learn.microsoft.com/en-us/azure/aks/use-system-pools

Maybe for the beginning helps others to understand the difference between the system vs user node pool.

azure/aks/modules/aks/main.tf
```diff
@@ -0,0 +1,62 @@
...
+  default_node_pool {
+    name                 = "system"
```
Langleu, line 16: production recommendation by Azure is to have 3 nodes for a system node pool if it's the only system node pool

azure/aks/modules/aks/main.tf
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    # Keep things simple for testing
+    max_pods                     = 30
+    only_critical_addons_enabled = true
+  }
+
+  # Network configuration
+  network_profile {
+    network_plugin    = "kubenet"
```
Langleu, line 36: just a suggestion to think about the default network plugin, also in regards to multi-region.
I didn't read into it but both option seem to have pro and cons and what do we recommend and why

azure/aks/modules/postgres-db/main.tf
```diff
@@ -0,0 +1,71 @@
...
+  zone = var.zone
+
+  high_availability {
+    mode                      = "ZoneRedundant"
+    standby_availability_zone = var.standby_availability_zone
+  }
+
+  # For testing purposes, we're enabling public access
+  # In production, you would restrict this and use private endpoints
+  public_network_access_enabled = true
```
Langleu, line 23: why do we need public network access?
As long as the resources within the vnet can communicate with each other, should worK?

azure/aks/modules/postgres-db/main.tf
```diff
@@ -0,0 +1,71 @@
...
+
+  # For testing simplicity, allow all IP addresses to connect
+  # NOT RECOMMENDED FOR PRODUCTION
+  lifecycle {
+    ignore_changes = [
+      # This allows automated tests to run without worrying about IP changes
+      tags["testing"],
+    ]
+  }
```
Langleu, line 32: I'm curious to learn more about this on how the lifecycle ignore relates to the IP topic.

azure/aks/modules/postgres-db/main.tf
```diff
@@ -0,0 +1,71 @@
...
+}
+
+# For testing purposes, allow all IPs to connect to the PostgreSQL server
+# NOT RECOMMENDED FOR PRODUCTION
+resource "azurerm_postgresql_flexible_server_firewall_rule" "allow_all" {
+  name             = "AllowAll"
+  server_id        = azurerm_postgresql_flexible_server.this.id
+  start_ip_address = "0.0.0.0"
+  end_ip_address   = "255.255.255.255"
+}
```
Langleu, line 42: shouldn't it be enough to allow the resources from the VNET access?

azure/aks/main.tf
Langleu, line 1: general feedback on the module usages, UX improvement worthy between variable and hardcoded values.

In EKS and ROSA for example local values were used and the files were split up in, where applicable,
`cluster`, `db`, `opensearch`, potentially `network`

Langleu (changes requested):
See comments that I've left.

Tried it out and works.
I'd suggest to concentrate for now on a single goal as I see a lot of `not recommended for production`, `just for testing`.

My suggestion concentrate on the MVP and similar UX as the other implementations.

azure/aks/main.tf
bkenez: Oh no, I forgot to push the commit where I had made that change.... sorry about that. 😔

azure/aks/main.tf
bkenez: Indeed, this was just for me to debug things while still developing, it will be removed once hooked up to automated tests

azure/aks/modules/network/main.tf
bkenez: Done

azure/aks/modules/aks/main.tf
bkenez: Added

azure/aks/modules/aks/main.tf
bkenez: Good point, I scaled it down, but it would be nice to have the production default recommendation here

azure/aks/modules/aks/main.tf
```diff
@@ -0,0 +1,62 @@
+resource "azurerm_kubernetes_cluster" "aks" {
```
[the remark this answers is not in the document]
bkenez: functionality added

azure/aks/variables.tf
```diff
@@ -0,0 +1,125 @@
...
+  description = "Kubernetes version to use for the AKS cluster"
+  type        = string
+  default     = "1.32.0"
+}
+
+# PostgreSQL Variables
+variable "postgres_version" {
+  description = "PostgreSQL version"
+  type        = string
+  default     = "16"
```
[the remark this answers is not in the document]
bkenez: Switched to 15

azure/aks/modules/aks/main.tf
bkenez: This is definitely an interesting topic, I chose kubenet out of familiarity, but might need to investigate if Azure CNI would be more appropriate depending on customer needs

azure/aks/modules/postgres-db/main.tf
bkenez: It should, but terraform cannot create the databases otherwise while I was testing locally, not sure what would be the best way to work around that yet

azure/aks/modules/postgres-db/main.tf
bkenez: That was a leftover comment copypaste accident while changing things in that block 🤦

azure/aks/modules/postgres-db/main.tf
bkenez: It should, once I can find a workaround for how to get terraform to connect to the DB for setting up the databases - see previous comment in this regard.

azure/aks/modules/postgres-db/main.tf
Langleu: as mentioned via PM to potentially approach it similar to how it's done in EKS to utilize a K8s job.
[ref](https://github.com/camunda/camunda-deployment-references/blob/main/aws/kubernetes/eks-single-region/setup-postgres-create-db.yml)

Same limitation there, I'd try avoiding to expose it unless really necessary.

azure/kubernetes/aks-single-region/variables.tf
```diff
@@ -16,6 +16,31 @@ variable "resource_prefix" {
   default     = "camunda"
 }
 
+variable "kubernetes_version" {
+  description = "Kubernetes version to use for the AKS cluster"
+  type        = string
+  default     = "1.28.5"
```
Langleu: would probably bump to a higher version as in e.g. EKS we're on 1.32 atm.

You can also think about how this should be updated via Renovate.

azure/kubernetes/aks-single-region/variables.tf
Langleu, line 1: In EKS / Openshift we don't double maintain the `variables.tf` and just consume the module directly.
So module has variables.tf and in the example implementation just specify the values directly and not add another layer.

Langleu: some values we see as "should be changed / adjusted by user intentionally" we have as local variables.

E.g. regions or clusternames, that sort of thing.

Langleu (changes requested):
left some smaller comments to potentially consider.
Thanks for the adjusted changes!
An architecture diagram (in our miro collection) could help to understand things better to get a better initial overview.

azure/aks/modules/postgres-db/main.tf
bkenez: Resolved, private endpoint and DB setup script implemented

azure/aks/modules/postgres-db/main.tf
bkenez: Resolved, private endpoint and DB setup script implemented

Langleu (changes requested):
Leaving the following items as feedback.
Please consider the items and what you want to move to the 2nd ticket or what you want to resolve now or dismiss completely.

Don't forget the availability zones topic, I don't see it listed in your list yet. (I added it manually first but removed it since I'm not sure it would otherwise maybe be overlooked as I saw another list in another ticket).

Langleu (approved):

Issue #124, opened 2025-03-06
Issues with the multibranch


leiicamundi:
- [ ] Schedules on branchs not available https://github.com/orgs/community/discussions/26374
- [ ] multiply number of spawned clusters for the same infrastructure

PR #128, opened 2025-03-10
feat(actions): internal scheduler

leiicamundi:
This PR address the issue related to GHA that does not allow scheduling workflows on dedicated branches other than main (https://github.com/camunda/camunda-deployment-references/issues/124).

The implementation creates a PR with a special branch and special name that should not be merged (draft).
This will trigger the schedule.

Example of a created PR https://github.com/camunda/camunda-deployment-references/pull/131 + associated run https://github.com/camunda/camunda-deployment-references/actions/runs/13774361651/job/38520133669

### TODO:
-  [x] in the scheduled workflows, include the env "IS_SCHEDULED" that either look at the github_event or the current branch name
- [x] ensure that each cluster bucket path uses the camunda version in the path
   - eks
   - ec2
   - rosa

### DIFF
--- a/.github/workflows/internal_global_branches_workflow_scheduler.yml
+++ b/.github/workflows/internal_global_branches_workflow_scheduler.yml
@@ -105,6 +105,7 @@ jobs:
 
     schedule-workflows:
         runs-on: ubuntu-latest
+        if: needs.define-matrix.outputs.schedules != '[]' && needs.define-matrix.outputs.schedules != ''
         needs:
             - define-matrix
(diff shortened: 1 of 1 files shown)


Langleu (changes requested):
Seems to work quite well 🚀.

The only thing I want you to consider is how the PRs are closed.
By having a separate branch, that means the workflow and state is executed from that branch. If the workflow doesn't finish within the hour and requires the state of the branch (checkout/v4) it will fail after the hour as the branch is gone.

PR #156, opened 2025-03-17
fix: always post diff if golden triggered

Langleu:
We should always execute the PR diff message otherwise you will miss out on it actually succeeding.

Related to - https://github.com/camunda/camunda-deployment-references/pull/144#issuecomment-2729069928

Had a meeting inbetween 😓

### DIFF
--- a/.github/workflows/aws_openshift_rosa_hcp_single_region_golden.yml
+++ b/.github/workflows/aws_openshift_rosa_hcp_single_region_golden.yml
@@ -108,5 +108,8 @@ jobs:
 
                   if [ -n "$DIFF_COMMENT" ]; then
                       echo "Deleting no diff comment with ID $DIFF_COMMENT"
-                      gh pr comment ${{ github.event.pull_request.number }} --delete --comment-id "$DIFF_COMMENT"
+                      gh api \
+                        --method DELETE \
+                        -H "Accept: application/vnd.github+json" \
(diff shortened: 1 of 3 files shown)


Langleu:
@leiicamundi your suggestion does not work as assumed.

Leaving it blank is equal to `No Diff`, which is the default.
The additions etc. would remain.

Langleu:
Adjusted the message slightly.

Langleu:
image

.github/workflows/aws_openshift_rosa_hcp_single_region_golden.yml
```diff
@@ -96,3 +96,22 @@ jobs:
                       ```sh
                       just regenerate-golden-file "${{ env.MODULE_DIR }}" "${{ env.S3_BUCKET_REGION }}" "${{ env.S3_BACKEND_BUCKET }}" "${{ env.S3_BUCKET_KEY }}"
                       ```
+                  comment-body-no-diff: ✅ No diff, you can ignore this comment and below footer.
+
+            - name: Delete specific comment containing the text and post diff
+              if: always() && github.event_name == 'pull_request'
```
Langleu, line 102: I didn't check all your changes but you could have it easier by using the output of the action

https://github.com/int128/diff-action

The above action would need a name but then you could just call it on `steps....outputs.different"

Langleu: talked about it, all good

PR #159, opened 2025-03-17
fix: openshift permissions for stable/8.6

leiicamundi:
Fix the permissions error with the following changes:
- Apply the permissions of cluster-admin before login 
- Use oc create project instead of just creating a namespace
- Fix ./procedure/gather-cluster-login-id.sh context 

We don't need the retry wrapper


Langleu (approved):
The changes seem to be run fine. Looked into the workflow run and after a couple retries seems to work.

PR #173, opened 2025-03-20
feat(rosa-dual-region): migrate dual region rosa

leiicamundi:
This PR introduces tests from the documentation instructions for the Dual region of openshift

The associated doc PR is https://github.com/camunda/camunda-docs/pull/5434

### DIFF
--- a/.github/actions/aws-openshift-rosa-hcp-single-region-cleanup/action.yml
+++ b/.github/actions/aws-openshift-rosa-hcp-single-region-cleanup/action.yml
@@ -99,5 +99,5 @@ runs:
                 ${{ inputs.max-age-hours-cluster }} ${{ inputs.target }} ${{ inputs.tf-bucket-key-prefix }}
 
               if [ "${{ inputs.delete-ghost-clusters }}" == "true" ]; then
-                ${{ github.action_path }}/scripts/cleanup-ghost-clusters.sh
+                ${{ github.action_path }}/scripts/cleanup-ghost-clusters.sh ${{ inputs.max-age-hours-cluster }}
               fi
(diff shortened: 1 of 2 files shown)


aws/openshift/rosa-hcp-dual-region/procedure/get-your-copy.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Download a copy of the reference architecture
+BRANCH="feature/migrate-dual-region-rosa"  # TODO: [release-duty] before the release, update this!
```
Langleu, line 4: don't forget to adjust after the merge as otherwise will fail.

generic/openshift/dual-region/procedure/chart-env.sh
```diff
@@ -30,4 +30,5 @@ export AWS_ES_BUCKET_REGION=""
 # The Helm release name used for installing Camunda 8 in both Kubernetes clusters
 export HELM_RELEASE_NAME=camunda
 # renovate: datasource=helm depName=camunda-platform registryUrl=https://helm.camunda.io
-export HELM_CHART_VERSION=11.3.0
+export HELM_CHART_VERSION="0.0.0-snapshot-alpha"
```
Langleu, line 33: you probably want to use `0.0.0-snapshot-latest` atm since `alpha` is 8.8 and likely not compatible.

Langleu (approved):
Had a quick look over it.
It's a lot and if something fails in the bigger picture, we can still fix it afterwards.

The test still seems to fail.
Even if ACM wasn't failing would assume it latest fails at the `alpha` version choice as it's 8.8 by now.

Langleu (approved):

PR #196, opened 2025-03-24
feat: golden files for everyone

leiicamundi:
### PR Summary  

This PR introduces several improvements:  

- **Golden files**: Adds golden files for all current and future modules to ensure consistency.  
- **ShellCheck Fix**: Removes the annoying ShellCheck `SC2155` warning.  
- **Skip Feature**: Implements a skip feature that allows disabling specific workflows in PRs using labels that match the corresponding file name.

### Note on bucket location  

Since we now have a single repository, the choice of a single bucket to manage all states seems obvious. However, not all workflows are aligned with this yet.  

I am currently using the bucket **tests-ra-aws-rosa-hcp-tf-state-eu-central-1**, which will likely need to be renamed later.  
The structure within this workflow follows the natural path of the module in this repository, making navigation easier.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -72,3 +72,5 @@ clusters-info:
 A label can be added to a pull request or issue to skip a specific workflow execution. For example, adding the label [`skip_aws_compute_ec2_single_region_tests`](https://github.com/camunda/camunda-deployment-references/labels/skip_aws_compute_ec2_single_region_tests) prevents unnecessary resource usage.
 
 If the required label does not exist, it must be created manually on GitHub at [GitHub Labels](https://github.com/camunda/camunda-deployment-references/issues/labels) with the color `#1D76DB`.
+
+_Note:_ One should create and apply the label during the creation of the PR; otherwise, the first run will trigger all workflows.


.github/workflows/aws_compute_ec2_single_region_golden.yml
```diff
@@ -25,15 +19,29 @@ env:
...
 
-    MODULE_DIR: ./aws/ec2/terraform/
+    MODULE_DIR: ./aws/compute/ec2-single-region/terraform/
 
 
 jobs:
+    triage:
+        runs-on: ubuntu-latest
+        outputs:
+            should_skip: ${{ steps.skip_check.outputs.should_skip }}
```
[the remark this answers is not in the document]
Langleu: nevermind, all good.

.github/actions/internal-triage-skip/action.yml
Langleu, line 1: important to note that it only matters if the label was set on PR creation or rerun.
Afterwards adding will still make it run.

.github/actions/internal-triage-skip/action.yml
```diff
@@ -0,0 +1,30 @@
...
+runs:
+    using: composite
+    steps:
+        - name: Get the current workflow filename and check for skip labels
+          id: check_labels
+          shell: bash
+          run: |
+              workflow_file_name=$(echo "${{ github.workflow_ref }}" | sed 's/@.*//')
+              workflow_file_name=$(basename "$workflow_file_name" | sed 's/\(.*\)\(\.yaml\|\.yml\)$/\1/')
+              echo "workflow_file_name=$workflow_file_name" | tee -a "$GITHUB_ENV"
```
Langleu, line 21: it's also part of the github context but can also be done your way.

`github.event.issue.labels.*.name`

.github/actions/internal-triage-skip/action.yml
Langleu, line 1: overall, okayish, would only do it in a public repo were we potentially don't have to pay for minutes run.
Wouldn't advice in private repos.

Similar could be achieved with a simple
```
if: !contains(github.event.issue.labels.*.name, "skip_blabla")
```

yes `skip_blabla` is hardcoded, there's no simple way to retrieve it but on the other hand how often does the filename change.

Nevertheless, your proposal is a valid approach.

.github/workflows/aws_compute_ec2_single_region_golden.yml
```diff
@@ -25,15 +19,29 @@ env:
...
+        steps:
+            - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
```
Langleu, line 35: not sure how hard you're set on this.
In theory you can skip the checkout step and it would simply default to the main branch for the action use.

Langleu: Maybe just to note as side info.

.github/actions/internal-triage-skip/action.yml
Langleu: sadly there's to this day not a serverless step that allows to create matrices / this kinda thing solely based on the context without having to spawn a whole machine.

.pre-commit-config.yaml
```diff
@@ -66,7 +67,7 @@ repos:
       rev: v0.9.0.6
       hooks:
           - id: shellcheck
-            args: [--external-sources]
+            args: [--exclude=SC2155, --external-sources]
```
Langleu, line 70: back at it again

image

DEVELOPER.md
```diff
@@ -41,3 +41,34 @@ Inside each workflow file, the `name` field is also standardized to maintain uni
...
+
+clusters-info:
+    needs:
+        - triage
+    if: needs.triage.outputs.should_skip == 'false'
+```
+
+### Creating and Using Skip Labels
+
+A label can be added to a pull request or issue to skip a specific workflow execution. For example, adding the label [`skip_aws_compute_ec2_single_region_tests`](https://github.com/camunda/camunda-deployment-references/labels/skip_aws_compute_ec2_single_region_tests) prevents unnecessary resource usage.
```
Langleu, line 72: maybe to add, that this should be done on PR creation already, otherwise the first run will always trigger.

Langleu: I can predict it will be forgotten regardless

aws/openshift/rosa-hcp-single-region/test/golden/golden.tfvars
Langleu, line 1: the eternal question, to file and empty file or not to file an empty file.
Up to you 🤷.

Langleu (changes requested):
Smaller description changes.
For the other stuff, think about it, consider and potentially dismiss, up to you.

I didn't know where it was going at the beginning of the PR but overall nice addition.
It's like 2 features in one. 😁

.github/workflows/aws_compute_ec2_single_region_golden.yml
Langleu: nope, nothing is required.

.github/actions/internal-triage-skip/action.yml
Langleu: just alternative way to access the labels.

.github/actions/internal-triage-skip/action.yml
Langleu: it depends on how quick you add them and whether the pipeline already ran the triage

Langleu (approved):

PR #202, opened 2025-03-26
chore: create aws cli action and enforce run pipefail

leiicamundi:
This PR extracts the current aws cli configuration in an internal reusable action.
It also add shell pipefail sanity for preventing silenced failures

### DIFF
--- a/.github/actions/aws-aurora-manage-cluster/action.yml
+++ b/.github/actions/aws-aurora-manage-cluster/action.yml
@@ -157,6 +157,9 @@ runs:
           run: |
               set -euo pipefail
 
+              # protect sensitive values
+              echo "::add-mask::${{ inputs.password }}"
+
               echo '${{ inputs.additional-terraform-vars }}' > /tmp/var.tfvars.json
               terraform plan -no-color -out aurora.plan \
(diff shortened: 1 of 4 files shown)


.github/actions/aws-configure-cli/action.yml
```diff
@@ -0,0 +1,45 @@
...
+inputs:
+    vault-addr:
+        description: The URL of the Vault server
+        required: true
+    vault-role-id:
+        description: The Vault Role ID
+        required: true
+    vault-secret-id:
+        description: The Vault Secret ID
+        required: true
```
Langleu, line 15: personal preference to simplify it by just passing in the secrets but this approach also works.

Langleu (approved):
The `x` can be helpful in case it fails and one just wants to see what was called / replicate it locally but also fine without the `x`.

AWS CLI extraction seems to work properly, the failing tests are not related to this change.

Langleu:
FYI: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#redacting-secrets-from-workflow-run-logs

PR #208, opened 2025-03-28
fix: use key prefix for single-region rosa

leiicamundi:
The current ROSA Single region state is not stored in the right place in the single bucket, this PR adds a prefix that matches with the RA


Langleu (approved):
it looks under `aws/openshift/rosa-hcp-single-region/` now 👍

PR #219, opened 2025-03-31
fix(aws): module vpc of eks don't have description on SG

leiicamundi:
Trivy requires security groups to have a description, this led to the failure of https://github.com/camunda/c8-multi-region/pull/541


Langleu (changes requested):
the test didn't like this change and will remember it.

Langleu (approved):

PR #230, opened 2025-04-04
tests(dual-region): dual region tests action

leiicamundi:
This PR introduces the dual-region tests of https://github.com/camunda/camunda-deployment-references/pull/173 directly in the existing action -> don't review it


Langleu:
@leiicamundi what exactly am I supposed to look here and how does it differ from #173 ?
The description says `don't review`.

PR #242, opened 2025-04-07
fix: cleanup ghost clusters of ROSA

leiicamundi:
This PR adds a cleanup for ghost clusters that are deleted by cloud nuke but still reported in the OpenShift cluster interface

Expected successful run: https://github.com/camunda/camunda-deployment-references/actions/runs/14306781873/job/40092387151?pr=242


Langleu (approved):
👻

Langleu (approved):
👍 , interesting that it requires the re-creation of the roles.

We had to manually cleanup some rosa clusters last week via the rosa cli but don't remember having to re-create roles and just triggered the delete.

PR #248, opened 2025-04-09
chore: golden file for eks

Langleu:
one jq transformation wasn't working properly, had to adjust it slightly.
The error prior didn't help much since it was just EOF.
Content is still the same, just slightly different sorted as before.


Langleu:
@leiicamundi , love the golden file implementation, very simple and generalized for everything.
Debugging was a bit tricky as the jq error just mentioned something generic at EOF, so had to delete parts to find the error but as we move forward with it for every environment, will eventually cover all edge cases.

PR #250, opened 2025-04-10
tests: integrate rosa domainless tests

leiicamundi:
This PR adds:

- [ ] domainless tests for ROSA Single Region
- [ ] uniformity with the matrix for ROSA single and dual region
- [ ] release name and namespace as variable for all archs
- [ ] a source as github env script

**Note:** some unit tests of the helm chart failed, nothing that is related to adaptation made for that. You can review it and I'll certainly have adjustments to apply later.

Related to https://github.com/camunda/team-infrastructure-experience/issues/614 and https://github.com/camunda/team-infrastructure-experience/issues/245

## TODO: 

- [x] relaunch tests on Tuesday
- [x] document new RELEASE NAME and NAMESPACE  (no more mention of HELM_RELEASE_NAME)
- [x] tested without skip_aws_openshift_rosa_hcp_single_region_tests (ip exhaustion)
- [x] tested without skip_aws_openshift_rosa_hcp_dual_region_tests (ip exhaustion)
- [x] backport to 8.7

### DIFF
--- a/.github/workflows/aws_modules_eks_rds_os_tests.yml
+++ b/.github/workflows/aws_modules_eks_rds_os_tests.yml
@@ -158,10 +158,12 @@ jobs:
               run: |
                   set -euo pipefail
 
-                  export TESTS_CLUSTER_ID="${{ needs.configure-tests.outputs.cluster_id }}"
-                  export TESTS_CLUSTER_REGION="${{ env.AWS_REGION }}"
-                  export TESTS_TF_BINARY_NAME="${{ env.TESTS_TF_BINARY_NAME }}"
                   just aws-tf-modules-test ${{ matrix.test_function }} "--junitfile ${{ matrix.test_function }}_unit-tests.xml"
+              env:
+                  TF_STATE_BUCKET_REGION: ${{ env.S3_BUCKET_REGION }}
+                  TESTS_CLUSTER_REGION: ${{ env.AWS_REGION }}
+                  TESTS_TF_BINARY_NAME: ${{ env.TESTS_TF_BINARY_NAME }}
+                  TESTS_CLUSTER_ID: ${{ needs.configure-tests.outputs.cluster_id }}
 
             # this is a workaround for test report not working as expected due to https://github.com/test-summary/action/issues/5
             - name: Filter logger.go from the test report (too large)


generic/openshift/dual-region/procedure/chart-env.sh
```diff
@@ -28,7 +28,9 @@ export AWS_ES_BUCKET_NAME=""
 export AWS_ES_BUCKET_REGION=""
 
 # The Helm release name used for installing Camunda 8 in both Kubernetes clusters
+export CAMUNDA_RELEASE_NAME=camunda
 export HELM_RELEASE_NAME=camunda
+
```
Langleu, line 33: what's the difference?

.github/scripts/gha-functions.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+#   You can source this file in any workflow step to make the functions
+#   available within that step.
+#
+# Usage:
+#   source .github/scripts/gha-functions.sh
+#   export_new_env_vars path/to/your/script.sh
+# ------------------------------------------------------------------------------
+
+export_new_env_vars() {
```
Langleu, line 16: could we add a description of what the purpose of this function is?

It's a bit hard to read, and I can imagine that looking back at this in a couple of weeks will require new deciphering.

In the end it adds variables that are currently not part of the env of the step as env for subsequent steps by adding those to the `GITHUB_ENV` as you probably rely on the new env vars somewhere and don't want to re-export them / having to think about it.

I think it might be slightly deceptive since env != GITHUB_ENV but that's probably not the purpose of this function.

.github/workflows/aws_openshift_rosa_hcp_single_region_tests.yml
```diff
@@ -443,12 +469,14 @@ jobs:
                   secrets: ${{ toJSON(secrets) }}
                   camunda-version: ${{ steps.camunda-version.outputs.CAMUNDA_VERSION }}
                   camunda-domain: ${{ env.CAMUNDA_DOMAIN }}
+                  camunda-domain-grpc: zeebe-${{ env.DOMAIN_NAME }}:443
```
Langleu, line 472: This will be breaking your tests.

For EKS I handled it similar to `env.CAMUNDA_DOMAIN` and exported it in the same if statement as otherwise I was overwriting the default and then the test will run against `zeebe-:443`.

So in case of `domain` it's defined otherwise empty.

Langleu (changes requested):
just some small comments from my side.
Overall looks quite good!
Strange branch history ^^.
Agree that some tests are failing unrelated to the changes of the branch.

Langleu (approved):
apart from the tests, consider this approved.
Don't forget your Todos 👍

PR #263, opened 2025-04-14
Outsource remaining scripts

Langleu:
Related to https://github.com/camunda/team-infrastructure-experience/issues/642

Ingress creation etc. was done as part of the test jobs already.
This pulls in the initial s3 bucket creation as well + a quick test. I haven't integrated it in the normal test as we don't make use of it due to a central bucket + this way we just solely focus on this use case.

There's a related docs PR - https://github.com/camunda/camunda-docs/pull/5559

After merge:
- [ ] backport to `stable/8.7`


Langleu:
I had it first in `generic` but didn't find the "generic" theme since it's aws specific.
Having a "generic" folder under aws felt counter intuitive to the generic generic folder.

Would you suggest we have a `common` folder?
```
common
    procedure
```

sort of thing?

Langleu:
@leiicamundi, in case you want to have a brief look over it again.
Moved it to `aws/common/procedure/s3-bucket` and renamed the test based on the folder structure.
There isn't anything specific for the s3-bucket so could use it for whatever in other docs.

PR #267, opened 2025-04-15
chore: rosa downgrade to 1 AZ to limit IP usage

Langleu:
related to https://github.com/camunda/team-infrastructure-experience/issues/539

Keeping it separate as for ROSA there's no perfect solution except downscaling the AZs as the TF module of ROSA has a hard dependency on the AZs to use for scaling the NAT Gateways. 

So it's just a proposal.


Langleu:
For now just want to see whether it's even possble ...
the VPC implementation of ROSA is quite restricted.

Langleu:
Closing this PR, may have to reconsider the approach.

One Test the subnet is `10.0.64.0/18` and another test it's `10.0.0.0/18` even though the calculation in the VPC module is the same and I'm supplying the same base range.

maybe utilizing `az_count` over `az` and let the module figure things out itself, could help ... not sure.

PR #269, opened 2025-04-15
tests: less renovate reruns

leiicamundi:
This PR fix the issue of renovate re-creating resources over the time. It works by adding a skip label when the run of renovate is successful.

An example successful run can be found here https://github.com/camunda/camunda-deployment-references/actions/runs/14476830878/job/40604413467

The only downside I've found is if renovate update in a row the same version, the second update will not get any tests but I think it's acceptable as we always have the scheduled runs as a safeguard.

**note:** I've paused renovate merges until this is merged, otherwise it just burns resources for nothing

### DIFF
--- a/.github/workflows/aws_openshift_rosa_hcp_single_region_tests.yml
+++ b/.github/workflows/aws_openshift_rosa_hcp_single_region_tests.yml
@@ -122,9 +122,7 @@ jobs:
 
                   echo "Filtering the matrix with strategy IS_SCHEDULE=$IS_SCHEDULE or IS_RENOVATE_PR=$IS_RENOVATE_PR"
                   if [[ "$IS_SCHEDULE" == "true" || "$IS_RENOVATE_PR" == "true" ]]; then
-                    # shellcheck disable=SC2086
-                    platform_matrix="$(yq '.matrix |= (.distro |= map(select(.schedule_only == true)))' \
-                      --indent=0 --output-format json ${CI_MATRIX_FILE})"
+                      echo "This PR is scheduled or coming from renovate, we test all scenarios without filtering."
                   else
                     # shellcheck disable=SC2086
                     platform_matrix="$(yq '.matrix |= (.distro |= map(select(.schedule_only == null or .schedule_only == false)))' \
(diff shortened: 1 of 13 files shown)


.github/workflows/aws_openshift_rosa_hcp_single_region_tests.yml
```diff
@@ -462,6 +462,12 @@ jobs:
                     kubectl -n "$TEST_NAMESPACE" logs "$pod_name";
                   done
 
+            - name: Prevent other runs for renovate
+              if: ${{ success() && env.IS_RENOVATE_PR == 'true' }}
+              env:
+                  GH_TOKEN: ${{ github.token }}
+              uses: ./.github/actions/internal-apply-skip-label
+
```
Langleu, line 470: the problem of calling it in matrix build is that one could succeed the other one fail and potentially hiding a problem with future runs as we're skipping it.
The extra comments in a PR don't matter that would be caused of it, don't have to overcomplicate it on first iteration.
Just saying that being called in a matrix may hide issues.

Langleu (changes requested):
not a deal breaker, just that `success()` is the default and wouldn't need to be called out but no problem in keeping it as is to emphasise the "success".

The requested changes is about the following:
See [comment](https://github.com/camunda/camunda-deployment-references/pull/269#discussion_r2046334737) concerning the matrix call, this may apply to other workflows as well (EKS single region for example)

.github/workflows/aws_openshift_rosa_hcp_single_region_tests.yml
Langleu: quick fix, could just be a separate job worst case, after cleanup or in between clean and integration, how you like.
The if would skip the execution on normal executions and only affects Renovate runs.

Langleu (approved):
wasn't first sure about the `report-success` `need` of `report-failure` but seems to work fine. Did a quick trial on a private repo and there's no "success" condition or similar required to run for the dependency as I wasn't sure how it behaves when the job is skipped.

PR #271, opened 2025-04-16
fix: schedule or renovate scenarios empty

leiicamundi:
This PR fix the case of renovate or shceduled PR, the  scenario was empty


Langleu (approved):
makes sense 👍 , good catch

PR #316, opened 2025-04-22
chore(cleanup): parallel destroy of tests

leiicamundi:
This PR adds parallelism to the destruction of the clusters, as destruction of a single cluster can take up to 40 minutes.

- Each destruction process is extracted in a sub-process
- The logs are stored in a dedicated file and streamed at the end
- A loop waits for all the sub-process to be completed

Example run; https://github.com/camunda/camunda-deployment-references/actions/runs/14595527242/job/40940391911?pr=316

With multiple clusters: https://github.com/camunda/camunda-deployment-references/actions/runs/14595255375/job/40939498446

### DIFF
--- a/.github/actions/aws-openshift-rosa-hcp-dual-region-cleanup/action.yml
+++ b/.github/actions/aws-openshift-rosa-hcp-dual-region-cleanup/action.yml
@@ -87,6 +87,7 @@ runs:
                 ${{ inputs.max-age-hours-cluster }} ${{ inputs.target }} ${{ inputs.tf-bucket-key-prefix }}
 
         - name: Upload cleanup logs
+          if: always()
           uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
           with:
               name: logs-${{ github.action }}-${{ github.job }}
--- a/.github/actions/aws-openshift-rosa-hcp-single-region-cleanup/action.yml
+++ b/.github/actions/aws-openshift-rosa-hcp-single-region-cleanup/action.yml
@@ -103,6 +103,7 @@ runs:
               fi
 
         - name: Upload cleanup logs
+          if: always()
           uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
           with:
               name: logs-${{ github.action }}-${{ github.job }}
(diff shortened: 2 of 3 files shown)


Langleu (changes requested):
Generally seems to work 👍.
I'd approached it differently by using a dynamic matrix and triggering it each matrix with the scoped key, but I'm good with the way it's done and we can keep it! Just showcasing a different approach. Please don't rework!

Only improvement I'd request is that the logs are archived in the github workflow, so I can download them afterwards and have a single view of the logs and don't have to mix and match the bits and pieces together.

.github/actions/aws-openshift-rosa-hcp-dual-region-cleanup/action.yml
```diff
@@ -85,3 +85,10 @@ runs:
               ${{ github.action_path }}/scripts/destroy-clusters.sh "${{ inputs.tf-bucket }}" \
                 ${{ github.action_path }}/../../../aws/openshift/rosa-hcp-dual-region/terraform/ /tmp/cleanup/ \
                 ${{ inputs.max-age-hours-cluster }} ${{ inputs.target }} ${{ inputs.tf-bucket-key-prefix }}
+
+        - name: Upload cleanup logs
```
Langleu, line 89: I'd suggest adding
`if: always()`

we do a check as part of the script that may exit 0 or 1 and would be nice to have the logs regardless.

.github/actions/aws-openshift-rosa-hcp-single-region-cleanup/action.yml
```diff
@@ -101,3 +101,10 @@ runs:
               if [ "${{ inputs.delete-ghost-clusters }}" == "true" ]; then
                 ${{ github.action_path }}/scripts/cleanup-ghost-clusters.sh ${{ inputs.max-age-hours-cluster }}
               fi
+
+        - name: Upload cleanup logs
```
Langleu, line 105: I'd suggest adding
`if: always()`

we do a check as part of the script that may exit 0 or 1 and would be nice to have the logs regardless.

Langleu (changes requested):
just one small addition from my side.

Langleu (approved):
🚀

PR #332, opened 2025-04-23
fix: add TF_STATE_BUCKET_REGION value for the s3 bucket

Langleu:
currently it was default to a different region (eu-west-2), while the suggested bucket is `eu-central-1`.
Resulting in the cleanup not kicking in since it was looking in the wrong bucket. 
And scheduled cleanup wouldn't find the resources either, till ultimately nightly cleanup would remove the resources.

- Added the env variable to set the bucket region correctly.

Should definitely help also with renovate as those resources would be blocking a couple of IPs as well till nightly nuke.

Will afterwards port to 8.7.

### DIFF
--- a/.github/workflows/aws_modules_eks_rds_os_tests.yml
+++ b/.github/workflows/aws_modules_eks_rds_os_tests.yml
@@ -158,10 +158,12 @@ jobs:
               run: |
                   set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


Langleu:
@bkenez this was related to the issue we saw earlier in the meeting where it mentioned `nothing found in the bucket`.
The bucket is really empty, since it always pushed into a different one.

PR #336, opened 2025-04-23
feat(openshift): private vpc and vpn ra alternative

leiicamundi:
Related to https://github.com/camunda/team-infrastructure-experience/issues/687

### This PR:
- adopts folder structure for modules of rosa hcp single region to have an optional vpn module (vpc-id cannot be determined in advanced, the split was required)
- adopts the cleanup script as the module now uses folders for modules (before merge, we need to ensure that no rosa single cluster is left, otherwise they will not be deleted)
- fix various inconsistencies in rosa dual region (outputs)
- integrates the vpn creation and configuration of the client in the create action
- the vpn certs and configs are stored in the s3 bucket
- TODO integrates a cleanup mechanism for vpn files in the s3 bucket in case they aren't deleted by terraform
- updates the golden file generate for rosa single region as we now have multiple modules
- in the workflow, the vpn config is managed similarly as the kubeconfig
- tests are unaffected as the DNS resolution is handled by the VPN
- VPN is in split mode: only the traffic of the cluster is sent trough the VPN
- the aws/modules/vpn can be reused in other implementations
- TODO: documentation explaining to the user how to connect to the vpn (for the reviewer, if you want to try, you can just download the config file and use it with any OpenVPN client, i.e. Tunnelbrick on mac)
- includes version agnostic links to the doc
- a flag controls if the test is supposed to use or not private-vpc

note: for future implementation, we should adopt the folder structure, otherwise adding a new module requires refactoring (heavy)

note 2: as this implementation does not reduces the number of eip consumed by the nat gateway, I'm wondering which strategy to adopt? 

## TODO
- [ ] doc PR refactoring links in the existing doc (8.7, 8.6, next)
- [ ] doc PR documenting usage of the VPN and vpc (8.7, next)
- [x] run a single test of dual region as I modified some files

## Scope
- [ ] EKS is not part of the scope of this initial PR, it will tackled later
- [ ] OpenShift dual region is **excluded** as it may requires 2 VPN (one for each region)

### DIFF
--- a/aws/openshift/rosa-hcp-single-region/terraform/vpn/test/golden/tfplan-golden.json
+++ b/aws/openshift/rosa-hcp-single-region/terraform/vpn/test/golden/tfplan-golden.json
@@ -83,14 +83,19 @@
           "name": "vpn_auth_rule",
           "provider_name": "registry.terraform.io/hashicorp/aws",
           "schema_version": 0,
-          "sensitive_values": {},
+          "sensitive_values": {
+            "timeouts": {}
+          },
           "type": "aws_ec2_client_vpn_authorization_rule",
           "values": {
             "access_group_id": null,
             "authorize_all_groups": true,
             "description": null,
             "target_network_cidr": "172.31.0.0/16",
-            "timeouts": null
+            "timeouts": {
+              "create": "10m",
+              "delete": "20m"
+            }
           }
         },
         "module.vpn.aws_ec2_client_vpn_endpoint.vpn": {


aws/modules/vpn/endpoint.tf
```diff
@@ -0,0 +1,97 @@
...
+
+resource "aws_security_group" "vpn" {
+  provider = aws.vpn
+
+  name_prefix = "client-vpn-endpoint-sg-${var.vpn_name}"
+  description = "Security group for Client VPN endpoint ${var.vpn_name}"
+  vpc_id      = var.vpc_id
+
+  ingress {
+    description = "Regulate ips allowed to access to the VPN"
```
Langleu, line 11:
```suggestion
    description = "Allow client VPN connections from approved IP ranges"
```

aws/modules/vpn/endpoint.tf
```diff
@@ -0,0 +1,97 @@
...
+    protocol    = "udp"
+    cidr_blocks = var.vpn_allowed_cidr_blocks
+  }
+
+  egress {
+    description = "Allow the VPN to access the internal network, unrestricted"
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
```
Langleu, line 23: Haven't looked into the details of the VPN config yet but just based on the security group would find the `cidr_block` and description a bit misleading.
`VPN to access the **internal network**` but cir_block `0.0.0.0/0`.

aws/modules/vpn/client-config.tf
Langleu, line 1: never a big fan of `local-exec`, it's like java dev describing how to do things locally but you run stuff in Kubernetes.

Nevertheless, small thought experiment for you:
Imagine running this in a remote environment (Terraform Cloud, CI system).
Barebones environment, you don't have AWS CLI installed on a managed environment.

aws/modules/vpn/generate-certificates.tf
Langleu, line 1: more `local-exec`.

My gut feeling would be to either have a third / fourth workspace that just takes care of creating and uploading the certificates ideally without the whole local exec stuff.
Or explain people on how to do provide the certificates (which format etc.) and expect them to be provided already to a path and have that as required variable.

Maybe a single workspace is also fine and I'm missing the point of all the `local-exec`. Maybe I'm missing documentation, not sure. Just looks confusing with upload / download via s3 etc. As said in the comment above, I don't see this running really outside of a local execution.

Then we could at least rely on something like the following:

```
data "aws_s3_object" "ca_cert" {
  bucket = "tests-ra-aws-rosa-hcp-tf-state-eu-central-1"
  key    = "aws/vpn-configs/hci-0ada35ce/ca-key.pem"
}

output "ca_cert" {
  value = data.aws_s3_object.ca_cert.body
}
```

The problem with the data source is it relies on something being already present. That's why my suggestion was the outsourcing.

aws/modules/vpn/import-certificates.tf
Langleu, line 1: Similar remarks to the other `local-exec` usages.

aws/openshift/rosa-hcp-single-region/terraform/vpn/config.tf
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+provider "aws" {
+  region = var.s3_bucket_region
+  alias  = "aws_bucket_provider"
+  default_tags {
+    tags = var.default_tags
+  }
+}
```
Langleu, line 29: maybe I'm missing the documentation part to fully grasp the dependency on S3.
Shouldn't it be possible to have the resulting files in the state as output?

Is the dependency coming from us for easy testing etc. or is it really required?

Langleu:
I leave it up to you!

PR #340, opened 2025-04-24
fix(tests): ensure error is thrown when object is not found

leiicamundi:
Instead of silencing the not found objects, we throw an error

### DIFF
--- a/.github/actions/aws-eks-cleanup-resources/action.yml
+++ b/.github/actions/aws-eks-cleanup-resources/action.yml
@@ -52,4 +52,4 @@ runs:
 
               ${{ github.action_path }}/scripts/destroy.sh "${{ inputs.s3-backend-bucket }}" ${{ github.action_path }}/../../../aws/modules/ \
                 "${{ inputs.temp-dir }}" ${{ inputs.max-age-hours }} ${{ inputs.target }} ${{ inputs.module-name }} ${{ inputs.camunda-version }} \
-                $([[ "${{ inputs.fail-on-not-found }}" == "true" ]] && echo "--fail-on-not-found"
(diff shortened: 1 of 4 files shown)


Langleu (commented):
wouldn't this cause nightly clean-ups to fail if nothing was found because nothing was present?
Giving us false positives.

.github/actions/aws-eks-cleanup-resources/scripts/destroy.sh
```diff
@@ -204,6 +204,11 @@ if [ "$ID_OR_ALL" == "all" ]; then
   resources=$(echo "$all_objects" | grep "/*.tfstate" | grep "$CAMUNDA_VERSION" | awk '{print $4}')
 else
   resources=$(echo "$all_objects" | grep "/*.tfstate" | grep "$CAMUNDA_VERSION" | grep "$ID_OR_ALL" | awk '{print $4}')
+
+  if [ -z "$resources" ]; then
```
Langleu, line 208: I think this part could be problematic - https://github.com/camunda/camunda-deployment-references/blob/main/.github/workflows/aws_modules_eks_rds_os_tests.yml#L212-L245

Essentially the golang tests will tear down the environment as part of the test.
So, in case everything succeeds, then the cleanup would fail since it wouldn't find anything.

I think the `always()` is for the case of abort job so maybe `failure() || cancelled()` could replace it.
So in case of failure we run the cleanup just in case, if it fails 🤷 , then was cleaned up properly prior.
Cancelled for gate keeping cancelled jobs.
Overall, ignoring success()

Langleu (commented):
see comment about an edge case

Langleu (approved):
nice one 👌

PR #348, opened 2025-04-25
chore: add eks focused cleanup job

Langleu:
Before it was cleaned up by the daily cleanup.
This is focused on the EKS ref arch and was missing so far.
Also makes it easier to trigger manually afterwards, similar to our other cleanups.
Will afterwards be backported to 8.7.

For simplicity, did not do a dynamic matrix but just cleanup all.

Example execution:
https://github.com/camunda/camunda-deployment-references/actions/runs/14666718799


Langleu:
the one you're reffering are the module tests, that we for now just migrated from the old repo as is.
They have their own daily cleanup.

These environments created as part of EKS single ref arch were cleaned up by the global cleanup.

PR #365, opened 2025-04-28
fix(tests): parallel tests matrix

leiicamundi:
This PR fixes the single region tests of OpenShift that are currently cancelled
https://github.com/camunda/camunda-deployment-references/actions/runs/14687083604/workflow

EDIT: in order to simulate multiple version, I tested it with RENOVATE flag, this is the reason why the bug wasn't spotted before (https://github.com/camunda/camunda-deployment-references/actions/runs/14704698042/job/41262179567)

## TODO:
- [ ] backport to stable/8.7


Langleu (approved):
nice one 👌 , job shows that all were properly executed now.

PR #387, opened 2025-05-02
feat(azure): add missing scripts and resources for Azure manual deployment

bkenez:
The scripts necessary for manual deployment, deploying the storage class, and adding an NSG that was supposed to be there (on the diagram).

Relates to: https://github.com/camunda/team-infrastructure-experience/issues/682
In order for the scripts to show up in the docs

Also relates to: https://github.com/camunda/team-infrastructure-experience/issues/681
Since the diagram includes things that are not actually in here (disregarding domain for now)


.github/workflows/azure_kubernetes_aks_single_region_golden.yml
```diff
@@ -89,6 +89,10 @@ jobs:
                   tenant-id: ${{ steps.secrets.outputs.AZURE_TENANT_ID }}
                   subscription-id: ${{ steps.secrets.outputs.AZURE_SUBSCRIPTION_ID }}
 
+            - name: Copy config_override.tf to module dir
+              run: |
+                  cp -r ./azure/kubernetes/aks-single-region/test/fixtures/fixture_config_override.tf ${{ env.MODULE_DIR }}
+
```
Langleu, line 95: as part of the script execution the following is done

```
  cd {{ justfile_directory() }}/{{ module_dir }}

  # Copy *.tf files from tests/fixtures/ to the current directory before running the plan
  if ls test/fixtures/fixture_*.tf 1> /dev/null 2>&1; then
    cp test/fixtures/fixture_*.tf ./
  fi
```

do we then need to explicitly copy the file here?
It will create eventually the case where one may wonder why we copy this file but not the other files etc.

.github/workflows/azure_common_procedure_storageaccount_test.yml
```diff
@@ -0,0 +1,191 @@
+---
+name: Tests - Azure Common Procedure Storage Account Creation Test
```
Langleu, line 2: Tests - Integration -

.github/workflows/azure_common_procedure_storageaccount_test.yml
```diff
@@ -0,0 +1,191 @@
...
+            AZURE_TF_KEY: camunda-terraform/terraform.tfstate
+
+        steps:
+            - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+              with:
+                  ref: ${{ github.ref }}
+                  fetch-depth: 0
+
+            - name: Install asdf tools with cache
+              uses: camunda/infraex-common-config/./.github/actions/asdf-install-tooling@14912e07221cd650bf9286e2a4733a138ca46374 # main
```
Langleu, line 69: typically we use the releases of our own common repo and not just main

Langleu (changes requested):
leaving some smaller comments

Langleu (approved):
didn't see anything left as part of the docs review part in regards to the code changes.

PR #508, opened 2025-05-27
fix: configure TLS for cluster membership of Tasklist/Operate

npepinpe:
# Description

This PR updates the documentation example to configure Tasklist/Operate for intra-cluster secured communication.

Related to: https://github.com/camunda/camunda-docs/issues/5869


npepinpe:
Note I didn't test it with Openshift itself, but tested with my own certs locally.

Langleu (approved):
From my side all good 👍, test was passing.
Could deploy it, Helm chart tests were green.
I've set up an environment on the side and had a look in the logs and it seems to work.

If there's nothing else from your side to look out for, would go ahead and merge + backport and ensure that the latest Helm charts are referenced to make it work due to the patch requirement of the applications.

npepinpe:
Sounds good, let's do it. I understood you will backport it?

PR #610, opened 2025-06-16
fix: only execute vpn export step when private is enabled

Langleu:
related to [slack](https://camunda.slack.com/archives/C076N4G1162/p1750048595082579) and [errored job](https://github.com/camunda/camunda-deployment-references/actions/runs/15671090483).

This PR fixes two issues in the ROSA single region test:
- in case of non private cluster (no VPN) usage, it failed a step that wasn't skipped
- in case of private cluster (VPN) usage, it failed as the tags were not sanitized in case of `[bot]` users.

- [x] I'll remove the overwrite commit around private before merging.

Run checking that the non VPN fix works:
- https://github.com/camunda/camunda-deployment-references/actions/runs/15674350221/job/44157027502?pr=610

Run that checks the VPN tag sanitization:
- https://github.com/camunda/camunda-deployment-references/actions/runs/15676228331?pr=610

### DIFF
--- a/.github/actions/aws-openshift-rosa-hcp-single-region-create/action.yml
+++ b/.github/actions/aws-openshift-rosa-hcp-single-region-create/action.yml
@@ -344,7 +344,15 @@ runs:
               cat vpn.tf
 
               raw_tags='${{ inputs.tags }}'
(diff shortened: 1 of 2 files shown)


Langleu:
I'll also add the missing sanitization to this.

Langleu:
@leiicamundi do you mind if we merge this already and keep the outsourcing of sanatization https://github.com/camunda/camunda-deployment-references/pull/611 to a separate PR?

PR #616, opened 2025-06-17
fix: rosa delete and re-create context

Langleu:
as well as skip awk in favour of k8s supported name returnal

related to https://camunda.slack.com/archives/C076N4G1162/p1750115840775489 and https://camunda.slack.com/archives/C076N4G1162/p1750140003725549 on slack.

I simplifies the `k8s get context` by removing the awk and just return names as is from the beginning.
I could imagine that maybe with the `awk` a space is in there and the `^$` won't work. `-x` option of grep is just an alias for `^$`.

Also I added removal of the old context just in case to allow recreating it, in case it still jumps into that line.

Backport:
- [ ] 8.6
- [ ] 8.7 if not possible the pending PR of VPN backport

### DIFF
--- a/.github/actions/aws-openshift-rosa-hcp-dual-region-create/action.yml
+++ b/.github/actions/aws-openshift-rosa-hcp-dual-region-create/action.yml
@@ -386,9 +386,11 @@ runs:
                   oc whoami
                   echo "Show existing contexts"
                   kubectl config get-contexts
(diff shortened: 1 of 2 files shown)


Langleu:
I can afterwards port it to 8.6 / 8.7.
Would assume it doesn't clash with your changes as you probably didn't touch it. If it does clash for 8.7 would add it to your PR but would let you know.

Langleu:
@leiicamundi , no problems backporting to 8.7 directly. Nothing changed code with the VPN setup it seems in those lines.

PR #646, opened 2025-06-24
fix: pin terraform versions

leiicamundi:
This PR fix a too loose versioning constraint on our terraform resources leading to unexpected major bumps

Context of this PR : https://camunda.slack.com/archives/C076N4G1162/p1750755961368559?thread_ts=1750728987.599229&cid=C076N4G1162

### TODO
- [x] investigate renovate update of those (https://docs.renovatebot.com/modules/manager/terraform/#range-constraints)
- [ ] backport stable/8.7
- [ ] backport stable/8.6

### DIFF
--- a/azure/modules/aks/providers.tf
+++ b/azure/modules/aks/providers.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "~> 3.90"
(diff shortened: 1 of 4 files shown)


Langleu (approved):
nice, now exactly the major bump would be covered and proposed as update. `>=` not ideal as it just says anything bigger than my current version that may clash with major upgrades.

PR #676, opened 2025-06-30
chore: output opensearch versions for renovate

Langleu:
related to https://github.com/camunda/team-infrastructure-experience/issues/783

similar to the rosa approach

--> https://camunda.github.io/camunda-deployment-references/opensearch_versions.txt


Langleu:
At least from the application side they're testing against the docker image of OpenSearch 2.19.x.

PR #768, opened 2025-07-24
chore: switch ec2 base to Ubuntu

Langleu:
related to https://github.com/camunda/team-infrastructure-experience/issues/772

Switches the ec2 base to Ubuntut o align with current Docker base image.
Probably doesn't change much considering it's debianesque, so the scripts work regardless of ubuntu or debian.

The volume topic required some minor adjustments but that could potentially work with newer debian amis as well apart from the user chowning.


Langleu:
@leiicamundi agree, but a question of prioritization from PM side.
If time allows in the end, I might be able to explore and add RHEL.

I really like the idea of podman but as you said yourself "not driven by the software itself but rather by IT constraints".

It might become an option further down the line, you just have to start somewhere.

PR #840, opened 2025-08-12
chore: move ec2 related items to generic/debian where applicable

Langleu:
The following just moves files to a new location `generic/debian` where applicable.
The PR does not add any functionality on top, just restructuring items where it made sense to me.
Similar to how we do things in the eks-single region, added symlinks.
Also got rid of the scripts folder and opted for `procedure` as well to align with the rest.


Langleu:
```
The Debian Cloud team publishes trixie for several cloud computing services:

Amazon EC2 (amd64 and arm64),
Microsoft Azure (amd64),
OpenStack (generic) (amd64, arm64, ppc64el),
PlainVM (amd64, arm64, ppc64el),
GenericCloud (arm64, amd64),
NoCloud (amd64, arm64, ppc64el)
```

Surprised to see that Azure doesn't have an arm image.

PR #844, opened 2025-08-13
chore: update retry action commit and add whole workflow option

Langleu:
adds the new rerun whole workflow option instead of just rerunning failed jobs.


Langleu:
While ACM wasn't at fault, it showcases that the full workflow retry is properly working.
I'll merge it after it has run through.

PR #878, opened 2025-08-20
fix(ci): use target branch instead of camunda version

leiicamundi:
This PR switch usage of the camunda version in favor of the target branch, it reduces (but do not limit in case of a breaking change between the PR and the target branch), the number of incompatibilities. Especially when main is targeting a camunda version that is also referenced in an other branch.

### TODO: 
- [ ] backport on all other branches
- [ ] enforce cleanup of leftovers using previous convention


Langleu (approved):

Langleu:
Within S3 `stable/8.7`
would then result in a folder
`stable` with `8.7`, `8.6` ...
which is fine 🤷

PR #908, opened 2025-08-25
fix: bump azure nodes to newer tier with higher network througput

Langleu:
The current Azure nodes example needs too long to pull the Camunda images.
Easily up to 20 minutes for a 500 mb image, with the default baseline being round about a minute.

Could also be local disk throughput issue, difficult to pinpoint.

The following proposal is switching to a newer tier. It has a higher network throughput and higher disk throughput, which hopefully helps with the image pull issues.

- [ ] backport 8.7

tried out instance types:
- Standard_D4s_v6 (latest) - not supported in all zones of a region
- Standard_D4as_v6 (latest, different type) - not enough quota, would need IT
- Standard_D4s_v5 (previous gen) - supported, should fit the quota, let's see how it perfoms ... 

Same problem for now, maybe not even a node related issue on their sizing but something wrong within AKS specific K8s version.
Or something from docker hub that they throttle to Microsoft but also strange considering that for some it works instant.


Langleu:
Closing for now as I don't know yet where the exact problem is.
One way to approach it, could be to mirror images to an interanl Azure registry to speed up the image pulls.

PR #915, opened 2025-08-27
feat(infra-deps): implements alternative supporting infrastructure deployment and 8.8 adjustments

leiicamundi:
Files supporting Vendor based methods: https://docs.google.com/document/d/1z7FVqc_YkQgYxM2X0NL1PsdBoaghaQdPxrKP-CCNwB4/edit?usp=sharing

### DIFF
--- a/.github/workflows-config/aws-openshift-rosa-hcp-single-region/test_matrix.yml
+++ b/.github/workflows-config/aws-openshift-rosa-hcp-single-region/test_matrix.yml
@@ -8,9 +8,7 @@ matrix:
         - name: OpenShift 4.19
           schedule_only: false
           # ! Unlike EKS, 'private_vpc' is defined in 'distro' and not in 'scenario'.
-          # private_vpc: true # latest always use vpn, olders use public
-          # TODO: revert
-          private_vpc: false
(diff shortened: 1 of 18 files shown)


generic/kubernetes/operator-based/keycloak/camunda-keycloak-no-domain-values.yml
```diff
@@ -0,0 +1,43 @@
...
+    identity:
+        auth:
+            enabled: true
+            type: KEYCLOAK
+            issuerBackendUrl: http://keycloak-service:8080/auth/realms/camunda-platform
+            # Local testing with port-forwarding
+            # using localhost:8081 -> keycloak-service:8080 (kubectl port-forward svc/keycloak-service 8081:8080)
+            issuer: http://localhost:8081/auth/realms/camunda-platform
+            publicIssuerUrl: http://localhost:8081/auth/realms/camunda-platform
+
```
Langleu, line 22: won't work with 8.8

From the road-to-8-8 work:

> The issuerBackendUrl is not supported atm by embedded Identity.
This means that publicIssuerUrl is used and must be accessible by both machines and users.
Only workaround for users is to do atm host file modifications

PR #917, opened 2025-08-27
feat(ai): enable github-copilot instructions

leiicamundi:
This PR enables copilot, this is just a copy/paste of the getting started in
https://docs.github.com/fr/copilot/how-tos/configure-custom-instructions/add-repository-instructions.

The goal is to have an automatic PR review made by copilot

## TODO

- [ ] backport 8.6
- [ ] backport 8.7


Langleu:
Anything worth to copy from https://github.com/camunda/camunda/blob/main/.github/copilot-instructions.md ?

Langleu:
btw. I don't think you followed the instructions outlined in https://docs.github.com/fr/copilot/how-tos/configure-custom-instructions/add-repository-instructions

You're supposed to run their agent with that query, which in turn creates a PR with the actual proposed changes targeting this instructions file.

PR #919, opened 2025-08-27
fix: Azure switch registry 8.7

Langleu:
backport of the azure fix: https://github.com/camunda/camunda-deployment-references/pull/916

will merge after successful run.

depends on https://github.com/camunda/camunda-deployment-references/pull/921

the helm chart reference is lagging too much behind as `allowInsecure` was only introduced in 12.3 or so.


Langleu:
the 5 mb postgres image may still cause issues which I find insane but for now will merge it so the majority of things will work.

PR #922, opened 2025-08-27
feat: single rule to regenerate all golden files

Langleu:
small improvement suggestion. Notice it sometimes as maintenance dri that I would like to just be able to run a single command that just regenerates all of it.

Just a suggestion.


Langleu:
Added your proposed change and fixed the pipefail topic 👍.
The arithmetic expansion was before triggering pipefail each time.

PR #1006, opened 2025-09-18
chore: make ec2 8.8 snapshot compatible

Langleu:
related to https://github.com/camunda/team-infrastructure-experience/issues/854

For now relying on snapshot due to all the config changes.
Have to rethink eventually the strategy for the upgrade tests but that's for future me to solve when we actually have "stable" versions.

I've cut away a lot of settings to rely more on the application defaults. Previous defaults were taken by the Helm Chart.
This changes implements:

- Support for connectors snapshot image usage
- adjust config changes to be compatible with the unified 8.8 config changes


Langleu:
can already be reviewed if wanted.
Otherwise wait till the blog changes in https://github.com/camunda/team-infrastructure-experience/issues/854 but they shouldn't impact the code here.

Langleu:
The alternative is to build your own `application.yaml` not much a fan of that. Environment variables will always approach the config, so it's always coming last, no matter what teams are brewing up.
Also the environment variable approach is nicer to repurpose for ECS since you can't mount an `application.yaml` in there and would have to bake it into the image or other weird approaches to get the file.

PR #1054, opened 2025-10-01
fix: use enterprise images and harbor registry for all deployments

leiicamundi:
This fixes the pipeline failure caused by unavailable Bitnami images from Docker Hub. The changes implement:

- Use values-enterprise.yaml as base for all Kubernetes deployments
- Apply registry-harbor.yml configuration for all scenarios (not just tests)
- Configure registry.camunda.cloud as the primary registry for better performance and reliability on Azure

The registry-harbor.yml is now applied unconditionally to avoid rate limits and availability issues with third-party registries.

For production deployments, users will need to create the registry-camunda-cloud secret as documented in:
https://docs.camunda.io/docs/next/self-managed/installation-methods/helm/configure/registry-and-images/install-bitnami-enterprise-images/#step-1-create-a-kubernetes-registry-secret

TODOs added for AWS EKS and OpenShift ROSA deployments to align with this new approach.

Fixes: https://github.com/camunda/camunda-deployment-references/actions/runs/18150271781

### DIFF
--- a/.github/workflows/azure_kubernetes_aks_single_region_tests.yml
+++ b/.github/workflows/azure_kubernetes_aks_single_region_tests.yml
@@ -88,7 +88,8 @@ env:
     TESTS_ENABLED: ${{ github.event.inputs.enable_tests || 'true' }}
     # TODO: [release-duty] before the release, update repo ref!
     # renovate: datasource=github-tags depName=camunda/camunda-platform-helm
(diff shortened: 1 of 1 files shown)


Langleu (commented):
instead of copying, loading, and merging.
Why not just add to the installation script

```
-f https://raw.githubusercontent.com/camunda/camunda-platform-helm/refs/heads/main/charts/camunda-platform-8.8/values-enterprise.yaml
```

It automatically affects all single region installation, provided the secret is present.
It would be in the docs and we separate that overlay from our configuration.
For a minor version the path will never change anyway since there's no patch values.

PR #1213, opened 2025-10-29
fix: add missing mapping rules for identity

leiicamundi:
TODO: backport to 8.8
Related to https://camunda.slack.com/archives/C09NXL6JE14/p1761593222696399

### DIFF
--- a/aws/kubernetes/eks-single-region-irsa/helm-values/values-domain.yml
+++ b/aws/kubernetes/eks-single-region-irsa/helm-values/values-domain.yml
@@ -159,11 +159,6 @@ orchestration:
                 admin:
                     users:
                         - admin
-                connectors:
-                    users:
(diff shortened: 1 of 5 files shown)


Langleu:
from my POV it's something that should be fixed in the default `values.yml` of the Helm Chart.

They're creating the default `admin` and `connectors` users and it should simply be part of the mapping over at them.
Otherwise the problem will remain for anyone that isn't using our refs.

Langleu (changes requested):
as mentioned, I see it as part of the defaults of the chart.

Langleu:
sure, but would still see it as part of the Chart itself as it only resolves the issue otherwise for a subset of users that follows our refs, while remaining a problem in the Chart itself that introduces said default users.

PR #1222, opened 2025-10-30
fix: mapping init

leiicamundi:
Related to https://camunda.slack.com/archives/C09NXL6JE14/p1761843100261479

simplifies usage & fix missing parts

upstream pr https://github.com/camunda/camunda-deployment-references/pull/1213

### DIFF
--- a/aws/kubernetes/eks-single-region-irsa/helm-values/values-domain.yml
+++ b/aws/kubernetes/eks-single-region-irsa/helm-values/values-domain.yml
@@ -159,11 +159,6 @@ orchestration:
                 admin:
                     users:
                         - admin
-                connectors:
-                    users:
-                        - connectors
-                    clients:
-                        - connectors
 
     ingress:
         grpc:
--- a/aws/kubernetes/eks-single-region-irsa/helm-values/values-no-domain.yml
+++ b/aws/kubernetes/eks-single-region-irsa/helm-values/values-no-domain.yml
@@ -161,11 +161,6 @@ orchestration:
                 admin:
                     users:
                         - admin
-                connectors:
-                    users:
-                        - connectors
-                    clients:
-                        - connectors
 
 console:
     enabled: false # by default, console is not enabled
--- a/aws/kubernetes/eks-single-region/helm-values/values-domain.yml
+++ b/aws/kubernetes/eks-single-region/helm-values/values-domain.yml
@@ -134,11 +134,7 @@ orchestration:
                 admin:
                     users:
                         - admin
-                connectors:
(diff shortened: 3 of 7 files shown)


aws/kubernetes/eks-single-region-irsa/helm-values/values-domain.yml
```diff
@@ -154,6 +154,16 @@ orchestration:
...
+        initialization:
+            defaultRoles:
+                admin:
+                    users:
+                        - admin
+                connectors:
+                    users:
+                        - connectors
+                    clients:
+                        - connectors
```
Langleu: I don't entirely get why the `connectors` have to be defined extra?

Looking at the default - https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.8/values.yaml#L2468-L2486

Then the only thing that really differs and is required is the admin part since the Helm Chart by default sets it to `demo` and we overwrite it to `admin`.

Langleu: image

Langleu (changes requested):
left a comment, let me know about it, otherwise makes sense to properly map the self-defined user to the admin role.
Could have been something in the helm chart if the firstUser is defined to be honest as people will run into this all the time. Maybe worth reporting to the Disro team and have it fixed upstream?

Langleu:
about the test app, we could have a test related overlay that pre-adds the required mapping as embedded identity does not check whether things exists, so anything can be pre-created already. May save the effort later on instead of utilizing APIs to add it.

Ideally Identity has documented how to properly manage the whole external IdP to embedded Identity usage.
Just checked, they do document things but it still requires a lot of mental effort from the user to make the connection between it.

aws/kubernetes/eks-single-region-irsa/helm-values/values-domain.yml
leiicamundi: > I don't entirely get why the connectors have to be defined extra? 

That's just for displaying it to the user - I'm fine removing it - but it's kind of obfuscated as it will be magically merged

Langleu: the magic begins with the user / app whatever it's in oidc mode

Langleu:
like a sync from IdP

Langleu (approved):
fine to merge it as is but as said, would look into getting it fixed upstream in the Chart instead.
Concerning the testing topic, as said could be added as an overlay and mappingRule

PR #1296, opened 2025-11-13
chore: dual-region openshift introduce multi-region tests

Langleu:
related to https://github.com/camunda/team-infrastructure-experience/issues/559
related to https://github.com/camunda/c8-multi-region/pull/827
- [x] Afterwards can change the target branch for the action.

integrating the dual-region tests for OpenShift.

The PR also fixes an issue I found along the way concerning the S3 implementation.

There's no TODO left in the changes, so ignore that workflow.

I left it as separate action, so whenever c8-multi-region is merged in here, we could just remove the git clone and adjust the working directory and it should work the same way.

There's a S3 cleanup now in the destroy action to ensure that the bucket and versions are all removed before the removal.


Langleu:
we should consider bumping the camunda version to 8.9 since stuff is gonna get differrent real soon.

Langleu:
Test finally passing - https://github.com/camunda/camunda-deployment-references/actions/runs/19363461317/job/55409158136?pr=1296

PR #1323, opened 2025-11-17
fix: use new identity client structure for extra venom client

Langleu:
related to [slack thread](https://camunda.slack.com/archives/C076N4G1162/p1763352230651699)

related to https://github.com/camunda/camunda-platform-helm/pull/4710

Uses the new client structure for Identity as otherwise we would get an index out of bound exception.
Alternative could have been to just decrease the index from 2 to 0


Langleu:
Merging to unblock the maintenance PRs as they also ran into the issue.

PR #1387, opened 2025-11-27
feat: add GH workflow to backport PRs to stable versions

marcel-dias:
## feat: add GH workflow to backport PRs to stable versions

To mark a PR as to be backport author should add labels to identify which branches should be updated.
example `backport stable/8.8` so the GH workflow will create a backport PR to the branch `stable/8.8`

The workflow has trigger when PR is merged or a comment starting with `/backport `  is added to the PR.

### DIFF
--- a/.github/workflows/internal_global_pr_backport.yml
+++ b/.github/workflows/internal_global_pr_backport.yml
@@ -2,56 +2,56 @@
 name: Internal - Global - Pull Request Backporting
 
 on:
-  pull_request:
-    types: [closed]
-  issue_comment:
-    types: [created]
+    pull_request:
+        types: [closed]
+    issue_comment:
+        types: [created]
 
 permissions:
-  contents: write # so it can comment
-  pull-requests: write # so it can create pull requests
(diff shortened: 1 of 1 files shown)


.github/workflows/internal_global_pr_backport.yml
```diff
@@ -0,0 +1,56 @@
...
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: write # so it can comment
+  pull-requests: write # so it can create pull requests
+
+jobs:
+  backport:
+    name: Create backport PRs
```
leiicamundi: You might need to adjust the usage here, the repo will enforce signed commit with the author.
To address this, you can setup the infraex bot to perform the commit (https://github.com/korthout/backport-action?tab=readme-ov-file#signing-cherry-picked-commits) 

I'm not sure of the current behavior, so let's first get this PR merged, then we test it and eventually adjust it with the following mechanic https://github.com/camunda/infraex-common-config/blob/main/.github/workflows/lint-global.yml

PR #1509, opened 2025-12-18
feat: ECS implementation part 1

Langleu:
related to https://github.com/camunda/team-infrastructure-experience/issues/970

Tests I'll outsource to a different ticket as the implementation itself is already plenty.
I'll still write some preliminary docs about this setup that I can later transform to camunda-docs.

Overall this introduces the base ECS setup for the expected alpha 3 scope of
- Orchestration Cluster
    - with RDBMS and IAM auth
- Connectors

Orchestration Cluster and Connectors house the ECS task definition services, and dependent configs and resources.
The root of the reference architecture houses a lot of the overarching resources that are shared between the ECS services.
One could put that into a separate module like `utilities` but this way the configuration is mostly opened as it should be owned by the customer.

More following tomorrow when I'm done with the preliminary docs.

// update: Added some [preliminary docs](https://github.com/camunda/camunda-deployment-references/blob/infraex-914-ecs-implementation/aws/containers/ecs-single-region-fargate/README.md). It's mostly just unstructured and a brain dump so I don't forget about it. Will eventually be converted to `camunda-docs`.

### DIFF
--- a/aws/containers/ecs-single-region-fargate/terraform/camunda.tf
+++ b/aws/containers/ecs-single-region-fargate/terraform/camunda.tf
@@ -15,15 +15,15 @@ module "orchestration_cluster" {
   ecs_task_execution_role_arn = aws_iam_role.ecs_task_execution.arn
 
   # Load Balancer configuration
-  alb_listener_http_80_arn   = aws_lb_listener.http_80.arn
-  alb_listener_http_9600_arn = aws_lb_listener.http_9600.arn
-  nlb_arn                    = aws_lb.grpc.arn
+  alb_listener_http_webapp_arn     = aws_lb_listener.http_webapp.arn
+  alb_listener_http_management_arn = aws_lb_listener.http_management.arn
+  nlb_arn                          = aws_lb.grpc.arn
 
-  enable_alb_http_80_listener_rule = true
+  enable_alb_http_webapp_listener_rule = true
   # management endpoint is unprotected, only enable if you know what you are doing.
   # Consider secure access alternatives via temporary jump host / VPN connected to VPC / lambda or step functions.
-  enable_alb_http_9600_listener_rule = false
-  enable_nlb_grpc_26500_listener     = true
+  enable_alb_http_management_listener_rule = false
+  enable_nlb_grpc_26500_listener           = true
 
   environment_variables = [
     {
@@ -128,15 +128,15 @@ module "orchestration_cluster" {
 module "connectors" {
   source = "../../../modules/ecs/fargate/connectors"
 
-  prefix                           = "${var.prefix}-oc1"
-  ecs_cluster_id                   = aws_ecs_cluster.ecs.id
-  vpc_id                           = module.vpc.vpc_id
-  vpc_private_subnets              = module.vpc.private_subnets
-  aws_region                       = data.aws_region.current.region
-  s2s_cloudmap_namespace           = module.orchestration_cluster.s2s_cloudmap_namespace
-  alb_listener_http_80_arn         = aws_lb_listener.http_80.arn
-  enable_alb_http_80_listener_rule = true
-  log_group_name                   = module.orchestration_cluster.log_group_name
+  prefix                               = "${var.prefix}-oc1"
+  ecs_cluster_id                       = aws_ecs_cluster.ecs.id
+  vpc_id                               = module.vpc.vpc_id
+  vpc_private_subnets                  = module.vpc.private_subnets
+  aws_region                           = data.aws_region.current.region
+  s2s_cloudmap_namespace               = module.orchestration_cluster.s2s_cloudmap_namespace
+  alb_listener_http_webapp_arn         = aws_lb_listener.http_webapp.arn
+  enable_alb_http_webapp_listener_rule = true
+  log_group_name                       = module.orchestration_cluster.log_group_name
 
   # IAM Roles (execution role centrally managed, task role module-specific)
   ecs_task_execution_role_arn = aws_iam_role.ecs_task_execution.arn
--- a/aws/containers/ecs-single-region-fargate/terraform/README.md
+++ b/aws/containers/ecs-single-region-fargate/terraform/README.md
@@ -34,8 +34,8 @@ This directory contains the Terraform implementation for the ECS single-region (
 | [aws_kms_key.secrets](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/kms_key) | resource |
 | [aws_lb.grpc](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb) | resource |
 | [aws_lb.main](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb) | resource |
-| [aws_lb_listener.http_80](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener) | resource |
-| [aws_lb_listener.http_9600](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener) | resource |
+| [aws_lb_listener.http_management](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener) | resource |
+| [aws_lb_listener.http_webapp](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener) | resource |
 | [aws_secretsmanager_secret.connectors_client_auth_password](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret) | resource |
 | [aws_secretsmanager_secret.db_admin_password](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret) | resource |
 | [aws_secretsmanager_secret.orchestration_admin_user_password](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret) | resource |
--- a/aws/containers/ecs-single-region-fargate/terraform/lb.tf
+++ b/aws/containers/ecs-single-region-fargate/terraform/lb.tf
@@ -10,12 +10,11 @@ resource "aws_lb" "main" {
   subnets = module.vpc.public_subnets
 }
 
-resource "aws_lb_listener" "http_80" {
+resource "aws_lb_listener" "http_webapp" {
   load_balancer_arn = aws_lb.main.arn
-  port              = "80"
+  port              = "80" # We use port 80 as example, feel free to change to 443 with proper SSL certs and domain
   protocol          = "HTTP"
 
-  # TODO: Maybe change default action
   default_action {
     type = "fixed-response"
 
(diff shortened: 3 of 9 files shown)


.gitignore
```diff
@@ -70,6 +70,7 @@ crash.*.log
 # control as they are data points which are potentially sensitive and subject
 # to change depending on the environment.
 *.tfvars.json
+*.tfvars
```
leiicamundi: might creates issues laters with

aws/modules/fixtures :

fixtures.default.aurora.tfvars
fixtures.default.eks.tfvars
fixtures.default.opensearch.tfvars

aws/containers/ecs-single-region-fargate/terraform/vpc.tf
```diff
@@ -0,0 +1,28 @@
...
+
+  name = local.name
+  cidr = local.vpc_cidr
+
+  azs             = local.azs
+  private_subnets = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 4, k)]
+  public_subnets  = [for k, v in local.azs : cidrsubnet(local.vpc_cidr, 4, k + 4)]
+
+  enable_nat_gateway     = true
+  single_nat_gateway     = true
```
leiicamundi: Any specific reason for us to create this SPOF?

aws/containers/ecs-single-region-fargate/terraform/config.tf
```diff
@@ -0,0 +1,28 @@
+terraform {
+  required_version = ">= 1.7.0"
+  backend "local" {
```
leiicamundi: Looks like it's the first time we use a local backend, any reason? not to enforce an s3 bucket?

aws/containers/ecs-single-region-fargate/terraform/camunda.tf
```diff
@@ -0,0 +1,190 @@
...
+  alb_listener_http_9600_arn = aws_lb_listener.http_9600.arn
+  nlb_arn                    = aws_lb.grpc.arn
+
+  enable_alb_http_80_listener_rule   = true
+  enable_alb_http_9600_listener_rule = true
+  enable_nlb_grpc_26500_listener     = true
+
+  environment_variables = [
+    {
+      name  = "CAMUNDA_CLUSTER_REPLICATIONFACTOR"
```
leiicamundi: I would recommend to extract this configuration in an external file (yaml or what you prefers) and directly consuming it here dynamically

aws/modules/ecs/fargate/orchestration-cluster/cloudwatch.tf
```diff
@@ -0,0 +1,13 @@
...
+  name              = "/ecs/${var.prefix}-camunda"
+  retention_in_days = var.cloudwatch_retention_days
+
+  tags = {
+    Name = "${var.prefix}-oc-log-group"
+  }
+}
+
+resource "aws_cloudwatch_log_stream" "orchestration_cluster_log_stream" {
+  name           = "${var.prefix}-orchestration-cluster-log-stream"
```
leiicamundi: just out of curiosity, do you plan any alarms? (cluster healtiness or other?)

aws/containers/ecs-single-region-fargate/terraform/config.tf
```diff
@@ -0,0 +1,28 @@
...
+    }
+  }
+}
+
+provider "aws" {
+  # set region via $AWS_REGION environment variable
+  region = "eu-west-1"
+
+  default_tags {
+    tags = {
```
leiicamundi: similarly to other implements, would be welcomed to have configurable tags

.gitignore
Langleu: good catch 👍.
Removed it again and transformed my local secret to a `tfvars.json`.

aws/containers/ecs-single-region-fargate/README.md
```diff
@@ -0,0 +1,189 @@
+# ECS single-region (Fargate) – Camunda 8 reference architecture
+
+This reference architecture deploys Camunda 8 (self-managed) to **AWS ECS on Fargate** in a **single region**, fronted by an **ALB** (HTTP) and **NLB** (gRPC), with **Aurora PostgreSQL** as RDBMS and supporting AWS primitives (EFS, S3, CloudWatch Logs, Secrets Manager, KMS).
+
+> ⚠️ This repository is intended for demonstration and learning purposes. This readme is temporary and will eventually be converted to camunda-docs.
+
+## What this stack includes
```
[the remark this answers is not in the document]
Langleu: not gonna stay, was just a brain dump before my FTO to have at least written out something already for others to consume and for me to come back to something from when it was still "fresh".
I'll transform that to `camunda-docs` in a different ticket.

aws/containers/ecs-single-region-fargate/terraform/config.tf
Langleu: not at all, was just for development purposes quicker.
Switched to s3 bucket / same definition we already use everywhere.

aws/containers/ecs-single-region-fargate/terraform/camunda.tf
Langleu: I used to do that in the poc to consume more or less my ec2 work but it felt cleaner / simpler for this example to just use them directly.
The "problem" with loading from yaml or similar is that you have to start templating again as well as transforming file to expected output, adding a bit of complexity on how this thing is overall managed.

Maybe I'll add an example as part of the documentation with some snippets.
What do you think?

aws/containers/ecs-single-region-fargate/terraform/camunda.tf
```diff
@@ -0,0 +1,190 @@
...
+  vpc_id              = module.vpc.vpc_id
+  vpc_private_subnets = module.vpc.private_subnets
+  aws_region          = data.aws_region.current.region
+
+  # IAM Roles (execution role centrally managed, task role module-specific)
+  ecs_task_execution_role_arn = aws_iam_role.ecs_task_execution.arn
+
+  # Load Balancer configuration
+  alb_listener_http_80_arn   = aws_lb_listener.http_80.arn
+  alb_listener_http_9600_arn = aws_lb_listener.http_9600.arn
```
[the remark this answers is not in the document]
Langleu: agree with the monitoring port, was mainly for development purposes.
I'll disable it by default and when enabled add a warning check about the usage of it.
In the docs I'll mention alternative ways to access it. Kinda downsides of ECS, no easy access to the machine without workarounds. We already expose an internal DNS record that could be consumed anywhere in the VPC, just requires a way to access it (VPN / Jump Host - EC2 / ECS Task / Lambda / Step functions / ...).
Or with the warning, temporarily enable it in the LB. 

Depends on how the user continues ALB usage with a real domain / certificate as well as on the monitoring port topic.

aws/modules/ecs/fargate/orchestration-cluster/cloudwatch.tf
Langleu: day 2 operations will be looked at after alpha3 till the release. Current target is to bring the base to ECS.
Depending on the outcome may be general applicable for ec2 as well or any other CloudWatch usage.

aws/containers/ecs-single-region-fargate/terraform/postgres.tf
```diff
@@ -0,0 +1,25 @@
+
+module "postgresql" {
+  source = "../../../modules/aurora"
+
+  # renovate: datasource=custom.aurora-pg-camunda depName=aurora-postgresql versioning=loose
+  engine_version             = "17.5"
```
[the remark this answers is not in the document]
Langleu: Yes, read about it as well but wanted to stick for now, also timeline wise, to what exists already within our domain.
Cost wise, for us when testing likely cheaper. For customers depends on how much load they have.
Since RDBMS overall is new for Camunda 8, would still for now stick to provisioned instances instead of serverless, also performance wise as we haven't even established the first one.
Would be careful for now to just establish it as a default.

aws/containers/ecs-single-region-fargate/terraform/ecs.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_ecs_cluster" "ecs" {
+  name = "${var.prefix}-cluster"
+
+  setting {
+    name  = "containerInsights"
+    value = "enabled"
```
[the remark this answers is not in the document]
Langleu: it is configurable to the user, it's not a module.
Same comment as about the nat gateway topic.

I don't think we have to now make everything as variable when it's not routed through a module or not reused somewhere.
Maybe in the future when there's let's say more alerts that kinda topic, there could be a variable that covers monitoring.

If you let's say, for the tests we want to disable it then agree, would be easier this way then overwriting it with tf overwrites.

Same for the Nat Gateway topic.

What do you think?

aws/containers/ecs-single-region-fargate/README.md
```diff
@@ -0,0 +1,189 @@
...
+- LoadBalancer
+- Secrets (+ KMS)
+
+The ECS related modules focus solely on their application specific configuration and usage with just covering the sole base required for ECS. Any additional required configuration should be managed by the user.
+
+For the Orchestration Cluster:
+
+- Creates IAM task role specific to this task access
+  - not to confuse with the task execution role. This is just scoped to what the task is allowed to access (S3 / Postgres)
+- Creates required S3 bucket for the node id provider
```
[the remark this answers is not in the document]
Langleu: not for the node-id provider one, it just creates unncessary costs.
You will have a million of versioned metadata file changes due to the lease topic, which makes no sense to keep.

aws/modules/ecs/fargate/connectors/ecs.tf
```diff
@@ -0,0 +1,127 @@
+
+resource "aws_ecs_task_definition" "connectors" {
+  family                   = "${var.prefix}-connectors"
+  execution_role_arn       = var.ecs_task_execution_role_arn
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
```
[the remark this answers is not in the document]
Langleu: it's in theory just setting two flags to enable `ARM` will have a look into it and exposing it through the module.
Funny as it may be, we still don't officially support `ARM` 😅 - at least docs only mention `amd64`

aws/modules/ecs/fargate/orchestration-cluster/templates/orchestration-cluster.json.tpl
```diff
@@ -0,0 +1,100 @@
...
+        "hostPort": 8080,
+        "protocol": "tcp",
+        "name": "rest"
+      }
+    ]
+  }
+  %{ if init_container_enabled ~}
+  ,
+  ${init_container_json}
+  %{ endif ~}
```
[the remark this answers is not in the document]
Langleu: It's a combination of two topics

https://developer.hashicorp.com/terraform/language/expressions/strings#string-templates

`%{ ... }` is a directive

`~` is whitespace stripping

but agree, the combintion can look a bit out of place 😁

Langleu:
Yes, CI will be added in a different work item, also to not overload this.

Dave tried it out yesterday as well and got it working on first try.
As well as we're using it for Zeebe in the benchmarks, though an older verison. Still have to update their definition to consume the latest changes.

aws/containers/ecs-single-region-fargate/terraform/postgres.tf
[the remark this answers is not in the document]
leiicamundi: i was thinking about tests and initial spawn time - nothing to act

leiicamundi (approved):
lgtm, thanks for the implementation of the changes and the explanations around various topic, great work!

PR #1604, opened 2026-01-05
fix: bump OpenShift to lowest supported version

Langleu:
4.16 reached EOL end of December

related to: https://camunda.slack.com/archives/C076N4G1162/p1767584666056779

TODO left on purpose.
After it works, will merge to unblock other topics.
Nice by Léo that the reference is in a single place, makes updating easy.

Got the 4.17 version from our renovate list that uses the cli to update it, so should be fine.


marcel-dias (approved):
LGTM

PR #1762, opened 2026-02-03
fix: fix max 50 labels limitation

leiicamundi:
This pull request enhances the internal triage skip GitHub Action to provide more flexible and user-friendly workflow skipping. Now, workflows can be skipped not only by adding specific labels, but also by including skip directives in the PR description, comments, or by checking boxes in an automatically managed checklist comment. Additionally, the action will post or update a checklist comment on pull requests, listing all available skip options for easier discoverability and use.

**Workflow skip logic improvements:**

* Expanded skip detection to check for skip directives in PR descriptions, comments, and checked items in the checklist comment, in addition to labels. This allows users to skip workflows through multiple convenient methods.
* Added support for skipping all tests by using the `testing-ci-not-necessary` directive in labels, PR text, comments, or checklist.

**User experience enhancements:**

* The action now automatically posts or updates a checklist comment on pull requests, listing all available `skip_*` labels as checkboxes. Users can check these boxes to skip specific workflows, making skip options more visible and easier to use.
* The checklist comment is managed to avoid overwriting user selections if it already exists, preserving checked boxes across updates.

* Updated the action description in `action.yml` to reflect the new skip mechanisms and checklist comment feature.

### DIFF
--- a/.github/actions/internal-triage-skip/action.yml
+++ b/.github/actions/internal-triage-skip/action.yml
@@ -69,13 +69,16 @@ runs:
               repo="${{ github.repository }}"
               checklist_marker="<!-- skip-workflow-checklist -->"
 
-              # Generate skip labels dynamically from workflow files
+              # Generate skip labels dynamically from workflow files that use this action
               echo "[debug] Generating skip labels from workflow files..."
               all_skip_labels=""
               for workflow_file in .github/workflows/*.yml .github/workflows/*.yaml; do
                 if [[ -f "$workflow_file" ]]; then
-                  workflow_name=$(basename "$workflow_file" | sed 's/\(.*\)\.\(yaml\|yml\)$/\1/')
-                  all_skip_labels+="skip_${workflow_name}"$'\n'
+                  # Only include workflows that use the internal-triage-skip action
+                  if grep -q "internal-triage-skip" "$workflow_file"; then
+                    workflow_name=$(basename "$workflow_file" | sed 's/\(.*\)\.\(yaml\|yml\)$/\1/')
+                    all_skip_labels+="skip_${workflow_name}"$'\n'
+                  fi
                 fi
               done
               # Remove trailing newline and sort


.github/actions/internal-triage-skip/action.yml
```diff
@@ -75,10 +159,56 @@ runs:
...
+              # Check PR description/body
+              echo "[debug] Checking PR description for skip directive..."
+              pr_body=$(gh pr view "$pr_number" --repo "$repo" --json body --jq '.body // ""')
+              if echo "$pr_body" | grep -qiE "$skip_pattern|testing-ci-not-necessary"; then
+                echo "[debug] Skip directive found in PR description"
+                echo "workflow_should_skip=true" | tee -a "$GITHUB_OUTPUT"
+                exit 0
+              fi
+
+              # Check PR comments for skip directive (excluding the auto-generated checklist comment)
```
Langleu: I'd probably remove all below here.
With comments you then have pagination again and potentially loose track of what you have ignored if you scatter them across comments.
Also GitHub truncates comments at some point, so you may miss one in the UI if you want to disable it.

Keeping it in the PR body only is probably better to have an overview. E.g. a default PR template could help to remind people about that option next to labels if a file name is too long.

Langleu (changes requested):
Leaving my pov / feedback.

Body description is a solution, bit unhandy but I don't think there's a good solution.
Either accept labels that are cut to 50 characters and do a matching to that (also a solution).
Or have people manually do the ignore in the PR body and maybe have a template that has all options preloaded without skip.

Example:

```
Skip workflows by adding `skip_` before the entry:

- aws_containers_ecs_single_region_fargate_tests
- aws_kubernetes_eks_single_region_tests
- aws_kubernetes_eks_single_region_irsa_tests
...
```

Could be an acceptable compromise next to labels.
Or as you did with a checkbox. Code wise just looked simpler to not overcomplicate it.

Langleu (changes requested):
tl;dr, makes sense, see my comment - https://github.com/camunda/camunda-deployment-references/pull/1762#discussion_r2763194958

Only thing to optimize from my pov is that we only list workflows that can be skipped.

Langleu (approved):

PR #1828, opened 2026-02-11
fix(cleanup): pre-nuke vpc of modules

leiicamundi:
related to https://camunda.slack.com/archives/C076N4G1162/p1770726563226269

This pull request improves the reliability and efficiency of the AWS Terraform cleanup process by proactively cleaning up VPC-associated resources before running `terraform destroy`. The main enhancement is the addition of pre-cleanup steps using `cloud-nuke` to remove orphaned AWS resources that are not tracked by Terraform, which helps prevent dependency errors and long timeouts during destroy operations.

**Enhancements to AWS resource cleanup:**

* Added a pre-cleanup step in `destroy_module()` for single-cluster destroys: now checks if a VPC matching the `group_id` exists and, if so, uses `cloud-nuke` to remove related resources before running `terraform destroy`. This prevents DependencyViolation errors caused by asynchronous cloud provider cleanup. (.github/actions/aws-generic-terraform-cleanup/scripts/destroy-resources.sh)
* Added a similar pre-cleanup step for dual-region destroys: checks for the existence of VPCs in both regions and uses `cloud-nuke` to clean up only those that exist, improving reliability and avoiding unnecessary errors or delays. (.github/actions/aws-generic-terraform-cleanup/scripts/destroy-resources.sh)


leiicamundi:
> Terraform rhcs 1.7.3 is broken, that's why 1.7.4 is out already. 

This makes sense for this week, but 1.7.3 does not explains the errors we had previous week

Regarding classic, I know, it's just AI :x

For the pipe || true it's expected, I mean, it's not fatal if it silently fails, tf delete then retry will take care of it

leiicamundi:
relaunching after the region cleanup

PR #1839, opened 2026-02-12
chore: change default CIDR block from /24 to /16

Langleu:
/24 is super small after splitting into subnets.


leiicamundi (approved):
lgtm as it aligns with eks, make sure to update the doc diagram if any

PR #2027, opened 2026-03-11
chore: bump ecs to latest alpha

Langleu:
didnt adjust renovate since sometimes there's alpha99 which may confuse it so kept it as a manual / optional thing atm


leiicamundi:
Hey @Langleu! Thanks for the PR ^^ I'll review it and look for automating that with renovate (we can exclude 99 in renovate regex)

PR #2187, opened 2026-03-31
feat(aws): enable EBS encryption by default in StorageClass

Langleu:
just a small suggestion.
It by default will encrypt the underlying volumes that are spawned through gp3-ebs.
For that it uses the default KMS key of an AWS account, can be adjusted but it's a cheap win.


leiicamundi:
closed by https://github.com/camunda/camunda-deployment-references/pull/2189

Thanks for your contribution @Langleu!

PR #2204, opened 2026-04-02
feat(helm-check): detect unknown keys in deployed Helm values

### LINKED ISSUE #2194: Non-existing value global.identity.enabled
According to the [default values](https://artifacthub.io/packages/helm/camunda/camunda-platform/13.5.3?modal=values&path=global.identity) there is no `global.identity.enabled` and it should probably be removed from the example:

https://github.com/camunda/camunda-deployment-references/blob/350b4401596a971be56d95d11274e0cafced293c/local/kubernetes/kind-single-region/helm-values/values-domain.yml#L14-L26

leiicamundi:
Add unknown keys validation to the internal-helm-deprecation-check action. When enabled (default: true), deployed values are checked against a strict version of the chart's JSON Schema to detect typos, removed, or renamed configuration properties that would be silently ignored at runtime.

The schema is automatically extracted from the deployed chart via helm pull. A Python script (stdlib only) recursively adds additionalProperties: false to all schema objects and reports any keys not defined in the schema.

Ref: https://github.com/camunda/camunda-platform-helm/issues/4564

close https://github.com/camunda/camunda-deployment-references/issues/2194


marcel-dias (commented):
I agree with the approach and setup. But it seems kinda expensive to run  a lot of workflows to validate changes in this action.  I would suggest to only validate in the local kind workflow. 

So removing `- .github/actions/internal-helm-deprecation-check/**` from other gh workflows.

leiicamundi:
>  I would suggest to only validate in the local kind workflow.

As each ref arch have its own value file, we need to validate all values files - the action is not supposed to evolve over the time

leiicamundi:
IRSA failure is ecpected, merging.

PR #2276, opened 2026-04-10
feat(eks): move ingress controller to Contour

### LINKED ISSUE #2103: [EKS Single Region] Migrate from ingress-nginx to Contour
Part of [#2102](https://github.com/camunda/camunda-deployment-references/issues/2102) — Epic: Migrate from ingress-nginx to Contour.

### Description

Replace ingress-nginx with [Contour](https://github.com/projectcontour/contour) in the EKS Single Region reference architecture.

**Current state**: ingress-nginx is installed via Helm with AWS NLB annotations.

### Files to modify

- [ ] `aws/kubernetes/eks-single-region/procedure/install-ingress-nginx.sh` → rename to `install-contour.sh`, replace with Contour Helm install using [AWS NLB guide](https://projectcontour.io/docs/1.31/guides/deploy-aws-nlb/)
- [ ] `aws/kubernetes/eks-single-region/helm-values/values-domain.yml` → update `ingressClassName` from `nginx` to `contour` (8.7/8.8), or switch to Gateway API values for 8.9+ (`global.gateway.enabled: true`, `global.gateway.className: contour`)
- [ ] `aws/kubernetes/eks-single-region/helm-values/values-no-domain.yml` → remove nginx-specific annotations if any
- [ ] Remove all `nginx.ingress.kubernetes.io/*` annotations from Helm values, replace with Contour equivalents if needed
- [ ] Update any README/documentation referencing ingress-nginx

### Current install script

```bash
helm upgrade --install \
  ingress-nginx ingress-nginx \
  --repo https://kubernetes.github.io/ingress-nginx \
  --version "$INGRESS_HELM_CHART_VERSION" \
  --set 'controller.service.annotations.service\.beta\.kubernetes\.io\/aws-load-balancer-backend-protocol=tcp' \
  --set 'controller.service.annotations.service\.beta\.kubernetes\.io\/aws-load-balancer-cross-zone-load-balancing-enabled=true' \
  --set 'controller.service.annotations.service\.beta\.kubernetes\.io\/aws-load-balancer-type=nlb' \
  --namespace ingress-nginx \
  --create-namespace
```

### Target Contour install

```bash
helm repo add contour https://projectcontour.github.io/helm-charts/
helm upgrade --install contour contour/contour \
  --namespace projectcontour \
  --create-namespace \
  --set envoy.service.annotations."service\.beta\.kubernetes\.io/aws-load-balancer-backend-protocol"=tcp \
  --set envoy.service.annotations."service\.beta\.kubernetes\.io/aws-load-balancer-cross-zone-load-balancing-enabled"=true \
  --set envoy.service.annotations."service\.beta\.kubernetes\.io/aws-load-balancer-type"=nlb
```

### Acceptance criteria

- [ ] Contour is installed via Helm with AWS NLB annotations
- [ ] Camunda is accessible via domain with TLS (cert-manager + Let's Encrypt)
- [ ] Zeebe gRPC is accessible via dedicated gRPC ingress/route
- [ ] external-dns creates Route53 records correctly
- [ ] CI tests pass

marcel-dias:
## feat(eks): move ingress controller to Contour"

- **feat(eks-sr): replace install-ingress-nginx.sh with install-contour.sh**
- **feat(eks-sr-irsa): update symlink to install-contour.sh**
- **feat(eks-sr): add Contour chart version and set ingressClassName**
- **ci: update ingress actions for Contour migration**
- **fix(eks-sr): add className to gRPC ingress and error handling**

resolves https://github.com/camunda/camunda-deployment-references/issues/2103
refs: https://github.com/camunda/team-infrastructure-experience/issues/933


marcel-dias:
@leiicamundi I believe this PR is ready to be reviewed, I checked the current errors and they do not seems related to the changes

leiicamundi:

---

## Review — `feat(eks): move ingress controller to Contour` (#2276 → `stable/8.9`)

Thanks for this — the migration is clean, well-scoped, and the opt-in design keeps every other ref-arch on nginx. I verified the value paths against the `camunda-platform-8.9` chart and the Contour `0.4.0` chart, and cross-checked the companion docs PR camunda/camunda-docs#8989. Overall this is in good shape; a few points below before approval.

### What I verified ✅

- **Helm value paths are valid** against `charts/camunda-platform-8.9/values.yaml`: `global.ingress.className`, `orchestration.ingress.grpc.className`, and `orchestration.service.annotations` are all documented parameters. The `orchestration.service.annotations.projectcontour.io/upstream-protocol.h2c` "unknown key" flagged by the deprecation-check bot is a **false positive** — `orchestration.service.annotations` accepts arbitrary annotations (it's a free-form map), and the same check also flags legitimate keys. Agreed with your assessment that the errors are unrelated.
- **Contour chart `0.4.0` exists** (appVersion 1.33.2) and its `values.yaml` sets `ingressClass.create: true` + `ingressClass.default: true`, so the chart creates an `IngressClass` named after the release (`contour`) and marks it cluster-default. `className: contour` will therefore bind correctly.
- **h2c annotation** (`projectcontour.io/upstream-protocol.h2c: '26500'`) is the correct mechanism to make Zeebe gRPC (cleartext HTTP/2) work through Contour/Envoy. Nice touch documenting that it's inert on OpenShift and ignored by other controllers.
- **`install-contour.sh`** improves on the old script: adds `set -euo pipefail` (the nginx script had no `set -e`) and uses `--set-string` for the NLB annotations.
- **Script-existence guard** in `aws-kubernetes-ingress-setup` is a good safety net.

### Questions / suggestions

1. **`ingressClass.default: true` — please add a documentation warning.** The Contour chart marks its `IngressClass` as the cluster default. On a fresh, dedicated ref-arch cluster this is harmless, but on an **existing** cluster it's a real footgun: any `Ingress` without an explicit `className` (for example resources previously served by an in-place nginx) would suddenly be claimed by Contour, which can break the running setup. Recommend a short **`> [!WARNING]`** note in the docs (and/or the install script) calling this out, so operators applying this to an existing cluster don't get surprised. This is documentation-only — no need to change the chart's default behavior.

2. **`nginx` is still the action default, but `install-ingress-nginx.sh` is deleted for `eks-single-region(-irsa)`.** The guard handles it (clear error, exit 1), but anyone invoking `aws-kubernetes-ingress-setup` with defaults against this ref-arch now fails. Since the EKS single-region ref-arch is fully migrated, was keeping `nginx` as the *input default* deliberate (to stay safe for Azure/other ref-archs that still ship the nginx script), or should this ref-arch's callers default to `contour`? A sentence in the action README clarifying the per-ref-arch availability would help.

3. **Removing the `global.opensearch.enabled` exclude-pattern + adding `continue-on-error` to the deprecation check.** This is a stable-branch-only relaxation (the `main` backport #2623 keeps it strict). The inline comment on `main` explains *why* `global.opensearch.enabled` is still needed (legacy Operate OpensearchConnector). Dropping the exclude here while making the check non-blocking is reasonable to unblock CI, but please add a short comment/TODO so a future reader knows the warning is knowingly tolerated and tracked.

### Merge sequencing ⚠️ (cross-repo)

The docs PR camunda/camunda-docs#8989 pulls `install-contour.sh` **live** from `blob/stable/8.9/...` via a `reference` code block. That file only exists on `stable/8.9` once **this** PR merges. So the order must be: **#2276 (this) merges first → then docs #8989**, otherwise the docs build 404s. The docs PR body already documents this dependency; just flagging it so the two don't merge out of order.

### Nits

- `install-contour.sh` is missing a trailing newline (minor; pre-commit may complain).

Once the intent on (1) and (2) is confirmed and the comment in (3) is added, this LGTM. 👍

aws/kubernetes/eks-single-region/helm-values/values-domain.yml
```diff
@@ -102,13 +103,18 @@ orchestration:
...
             host: zeebe-${CAMUNDA_DOMAIN}
             tls:
                 enabled: true
                 secretName: zeebe-c8-tls-grpc
             annotations:
                 kubernetes.io/tls-acme: 'true'
 
+    service:
+        annotations:
+            projectcontour.io/upstream-protocol.h2c: '26500'
```
leiicamundi, line 116: ✅ Verified `orchestration.service.annotations` is a valid, documented key in `camunda-platform-8.9` (free-form map), so the deprecation bot's “unknown key” flag on this annotation is a false positive. h2c on 26500 is the correct mechanism for Zeebe gRPC through Contour/Envoy.

aws/kubernetes/eks-single-region/procedure/install-contour.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+helm upgrade --install \
+  contour contour \
+  --repo https://projectcontour.github.io/helm-charts/ \
```
leiicamundi, line 6: Contour chart `0.4.0` sets `ingressClass.{create:true,default:true}`, so it creates a **cluster-default** `IngressClass contour`. On an **existing** cluster this can hijack any `Ingress` without an explicit `className` and break the running setup. Recommend a `> [!WARNING]` in the docs about applying this to an existing cluster (docs-only — no chart change needed).

.github/workflows/aws_kubernetes_eks_single_region_tests.yml
```diff
@@ -874,15 +874,11 @@ jobs:
                   ./generic/kubernetes/single-region/procedure/install-chart.sh
 
             - name: 🔍 Check for Helm deprecation warnings
+              continue-on-error: true
```
leiicamundi: Making the deprecation check non-blocking + dropping the `global.opensearch.enabled` exclude is a stable-only relaxation (the `main` backport #2623 keeps it strict). Reasonable to unblock CI, but please add a short comment/TODO so the tolerated warning stays tracked.

leiicamundi (commented):

aws/kubernetes/eks-single-region/procedure/install-contour.sh
marcel-dias: I will modify the ingress class.default to false and avoid any recommendation on migration because the reference architecture is about greenfield, not migration. If we want to tackle migration, we should go for a blog post.

PR #2277, opened 2026-04-10
feat: ECS dual-region Fargate with configurable networking, storage, and guided deployment commands

marcel-dias:
## Summary

Active-active dual-region ECS Fargate deployment for Camunda 8 with a 3-state Terraform layout (vpc → infra → app via `terraform_remote_state`), configurable networking (Transit Gateway or VPC Peering), configurable secondary storage (Aurora Global or OpenSearch), BYO-VPC support, and a 6-step skill-driven deployment workflow.

## Layout — 3 Terraform states

Customers can adopt only the layers they need.

- `terraform/vpc/` — VPCs (greenfield), cross-region peering or Transit Gateway, optional Route 53 Resolver. Supports **BYO-VPC** for customers with existing networking; `check` blocks enforce the contract at plan time. Slow but rarely changes.
- `terraform/infra/` — Aurora Global / OpenSearch, ECS clusters, ALB + NLBs, KMS, S3, secrets, IAM. Reads vpc outputs via `terraform_remote_state`. ~15-25 min.
- `terraform/app/` — Camunda orchestration-cluster + connectors task definitions and services. Reads infra outputs. ~30 s.

A legacy combined state at `terraform/clusters/` remains for back-compat but is not used by the new workflow.

### Reference architecture
- `aws/containers/ecs-dual-region-fargate/` — full dual-region ECS Fargate setup
- 8 Zeebe brokers (4 per region), replication factor 4, 8 partitions, **zone-aware partition distribution** so partitions span both regions
- Cross-region Raft consensus via internal NLB + Transit Gateway or VPC Peering
- Connectors (1 per region) with Service Connect for local gateway access
- One-time `db_seed` task per region (RDBMS) creates `camunda` IAM DB user

### New Terraform modules
- `aws/modules/aurora-global/` — Aurora Global Database (writer + cross-region reader, KMS per region, IAM auth, 30s `wait_for_primary` so the secondary actually joins, expanded `ignore_changes` so it stays joined)
- `aws/modules/transit-gateway/` — TGW peering for cross-region VPC connectivity

### Configurable infrastructure
- **`networking_mode`** (`transit_gateway` | `vpc_peering`)
- **`secondary_storage_type`** (`rdbms` | `opensearch`)
- **`byo_vpc`** — supply existing VPC IDs, subnets, CIDRs, route tables
- **`region_0` / `region_1`** — any pair of AWS regions
- **`enable_cross_region_dns_resolver`** — optional Route 53 Resolver for Cloud Map DNS forwarding
- All user inputs persisted as `terraform.tfvars`

### Claude Code skills (`/ecs-dual-region/*`)
Interactive guided 6-step deployment with Camunda platform context:

1. `1-configure` — collect inputs (regions, networking, storage, BYO-VPC), write three tfvars files
2. `2-deploy-vpc` — apply vpc state (greenfield or BYO)
3. `3-deploy-infra` — apply infra state, reads vpc remote state, ~15-25 min for Aurora Global
4. `4-deploy-camunda` — apply app state, wait for ECS steady state + Raft quorum
5. `5-verify` — PASS/FAIL health checks (ECS services, Zeebe topology, Aurora/OpenSearch)
6. `6-cleanup` — ordered teardown (app → infra → vpc)

### Tests
- Terratest suites for the 4 new aws/modules
- Greenfield end-to-end Terratest covering vpc → infra → app
- BYO-VPC + failover/failback Terratest

### Procedure scripts
- `verify_dual_region.sh` — full health check
- `failover.sh` / `failback.sh` — planned/unplanned region failover procedures
- `test_cross_region_dns.sh` — cross-region DNS resolution sanity check

## Camunda 8.10 compatibility & bugs found during validation

These commits land fixes for issues surfaced during a fresh end-to-end deploy of this branch on a test environment:

- **`fix(ecs-dual-region): separate connectors image and raise ECS service timeouts`** — connectors ship as `camunda/connectors-bundle` on docker.io (public), not `camunda/camunda`. Passing registry.camunda.cloud creds to a docker.io image confuses ECS task pulls. Also raises Fargate `service_timeouts` from 10m to 30m to accommodate dual-region cold start + Raft formation.
- **`fix(modules/ecs-orchestration-cluster): scope S3 IAM policy by region and require TLS`** — adds `aws:RequestedRegion` + `aws:SecureTransport=true` conditions to the orchestration-cluster S3 access policy. KMS access intentionally keeps `kms:ViaService` as a wildcard (`s3.*.amazonaws.com`) because S3 occasionally routes KMS calls through unexpected regional endpoints (notably with bucket-key optimization).
- **`fix(ecs-dual-region): rename partitioning config keys to ZONEAWARE for Camunda 8.10`** — `CAMUNDA_CLUSTER_PARTITIONING_REGIONAWARE_REGIONS_*` was renamed to `CAMUNDA_CLUSTER_PARTITIONING_ZONEAWARE_ZONES_*`. Without this rename, brokers fall back to no-zone-awareness and partitions concentrate without honoring region boundaries.
- **`fix(ecs-dual-region): allow cross-region Zeebe command API on camunda-ports SG`** — the cross-region rule only allowed 26502 (Raft). Broker-to-broker command forwarding on 26501 and gateway gRPC on 26500 were blocked, so brokers registered cross-region but every command timed out — 8 brokers visible, 0 leaders elected. Widened the cross-region range to 26500-26502.
- **`fix(modules/aurora-global): expand secondary ignore_changes to keep it attached to global`** — when an Aurora Global secondary is attached, AWS auto-populates `replication_source_identifier` with the primary's ARN. The provider treats this as configured-empty on every plan and clears it on apply, which AWS translates into `PromoteReadReplicaDBCluster` — detaching the secondary. This was observed in production: a plan that only touched two security groups detached the secondary via a stealth `~ replication_source_identifier: arn -> null` change. Adds `replication_source_identifier` and `engine_version` to `ignore_changes` alongside the existing `global_cluster_identifier`. Mirrors the upstream `terraform-aws-modules/terraform-aws-rds-aurora` pattern.
- **`fix(ecs-dual-region): enable Basic auth on orchestration cluster for 8.10`** — sets `CAMUNDA_SECURITY_AUTHENTICATION_METHOD=basic` on the orchestration-cluster task. Without it, the broker bootstraps users but never wires up a Spring Security filter chain — any `/v2/*` request fails with `CamundaAuthenticationException: ... Spring Authentication 'null'`. The Helm chart sets this automatically; raw-container deployments didn't surface the requirement before 8.10 because the endpoints connectors call were on separate, looser-auth services. 8.10's unified API enforces auth on all of them.

## Architecture

```
Region 0                                Region 1
┌─────────────────────┐                 ┌─────────────────────┐
│  ECS Cluster        │                 │  ECS Cluster        │
│  ├─ 4 Zeebe brokers │◄── Raft ──────►│  ├─ 4 Zeebe brokers │
│  └─ 1 Connector     │   (NLB 26502)  │  └─ 1 Connector     │
│                     │                 │                     │
│  ALB (8080/9600)    │                 │  ALB (8080/9600)    │
│  NLB (gRPC 26500)   │                 │  NLB (gRPC 26500)   │
│                     │                 │                     │
│  Aurora Writer ─────│── async repl ──►│  Aurora Reader      │
│  S3 Backup Bucket   │                 │  S3 Backup Bucket   │
└────────┬────────────┘                 └────────┬────────────┘
         │          Transit Gateway              │
         └──────── or VPC Peering ───────────────┘
```

## Test plan

- [x] `terraform validate` passes for vpc, infra, app
- [x] Greenfield end-to-end deploy on a real account (test-4 on us-east-1 + us-east-2): vpc + infra + app all apply, brokers reach 8/8 with 8 partition leaders, both ALBs serve `/v2/topology`
- [x] Cross-region command API reachable (broker-to-broker forwarding works after SG fix)
- [x] Aurora Global secondary stays attached across unrelated applies (after `ignore_changes` fix)
- [x] Orchestration cluster REST API accepts Basic auth on `/v2/process-definitions/search` (after auth-method fix); connectors importer succeeds
- [x] Terratest module suites green
- [ ] OpenSearch + VPC Peering combination still validates (no end-to-end run yet)
- [ ] BYO-VPC deploy end-to-end on a real account (covered by Terratest only)
- [ ] `terraform destroy` cleans up all resources


leiicamundi:

---

## Review — `feat: ECS dual-region Fargate with configurable networking, storage, and guided deployment` (#2277, draft → `main`)

This is an impressive, ambitious piece of work — a full active-active dual-region ECS Fargate reference with Aurora Global Database, configurable TGW/peering networking, and a guided deployment flow. The `aws/modules/aurora-global` module in particular is thoughtfully engineered. Since it's a large draft (+11,662 / 79 files) and currently conflicting, here's a structured review to help shape it toward merge. I've grouped findings by severity.

---

### 🔴 Blocking

**1. Security regression in the shared `orchestration-cluster` module S3 policy.**
`aws/modules/ecs/fargate/orchestration-cluster/s3.tf` removes three IAM policy `Condition` blocks:
- `aws:RequestedRegion` (on bucket + object statements)
- `aws:SecureTransport = true` (TLS-only enforcement on object access)
- `kms:ViaService = s3.<region>.amazonaws.com`

This module is **shared** — it's also consumed by the existing single-region reference `aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf` (`source = ../../../../modules/ecs/fargate/orchestration-cluster`). On `main` that module still carries all three conditions, so this change **weakens the IAM posture of single-region too**, not just the new dual-region ref-arch. Dropping `aws:SecureTransport` especially is a real regression (it removes the TLS-only guard, an OWASP-relevant control).

The PR description says these conditions "caused 403 errors." That's almost certainly the **cross-region** access pattern (a region-1 task reaching the region-0 bucket/KMS makes `RequestedRegion`/`ViaService` mismatch). The fix should *re-scope*, not *remove*:
- For `aws:RequestedRegion` / `kms:ViaService`: allow the set of regions actually in play (e.g. a list of both region values) rather than deleting the condition.
- For `aws:SecureTransport`: please **keep** it — it shouldn't cause 403s on legitimate TLS traffic, and removing it is a security downgrade.

If the cross-region coupling makes per-region conditions awkward, an alternative is to keep the strict conditions in the single-region module and parameterize the relaxed/multi-region variant behind a variable so single-region stays locked down by default.

---

### 🟠 Architecture / maintainability

**2. Three Terraform roots with heavy duplication — please consolidate to one model.**
The ref-arch ships three roots:
- `terraform/clusters/` — monolithic; this is what the `README.md` and **all five** `.claude/commands/ecs-dual-region/*` drive.
- `terraform/infra/` + `terraform/app/` — a two-phase split (`app` reads `infra` via `terraform_remote_state`, local backend).

I diffed them: **13 of the `.tf` files are byte-identical** between `clusters/` and `infra/`, and `clusters/` is effectively `infra/` + `camunda.tf` + `iam.tf`. The `infra/`+`app/` pair is **not referenced** by the README or any command, so it reads as dead/duplicate code that will drift. Recommend picking **one** layout (the monolithic `clusters/` that everything already uses, or the `infra/`+`app/` split if two-phase is the goal) and deleting the other. As-is, every future change has to be made in two places.

**3. `clusters/` vs the rest of the repo naming.** Other ECS refs use `terraform/cluster/` (singular) for the single root. Worth aligning the directory name with the established convention once the layout is settled.

---

### 🟡 Repo conventions / style

**4. Root `CLAUDE.md` overwrite conflicts with the established convention.**
`main` intentionally keeps a 7-line `CLAUDE.md` that delegates: *"Read `AGENTS.md` for commands, rules, and architecture."* This PR replaces it with a 361-line standalone file. That's both the source of a merge conflict and a divergence from the AGENTS.md-as-single-source convention. Suggest either folding the genuinely new content into `AGENTS.md` (and keeping the stub), or, if a longer `CLAUDE.md` is wanted, coordinate that as its own change rather than inside this feature PR.

**5. Internal planning docs are being committed to the repo.**
`docs/brainstorms/…`, `docs/plans/…`, and `docs/superpowers/{plans,specs}/…` (≈2,700 lines across 6 files) are **not present on `main` and not gitignored**. These look like working/brainstorm artifacts rather than reference documentation. Recommend removing them from the PR (or adding the directories to `.gitignore`) so the repo stays focused on the reference architecture itself.

**6. Committing `.claude/commands/`.** The repo's `.gitignore` only ignores `.claude/scheduled_tasks.lock` and `.claude/settings.local.json`, so committing slash-commands under `.claude/commands/` is a new pattern for this repo. Not wrong, but a maintainer-level decision — worth calling out explicitly so it's an intentional choice. (If kept, the commands are well written.)

---

### 🟢 What's good (please keep)

- **`aws/modules/aurora-global`** is excellent: the `time_sleep.wait_for_primary` ordering with the explanatory comment, `ignore_changes = [global_cluster_identifier]` to avoid the provider's detach/re-add trap, `storage_encrypted` + per-region KMS with `enable_key_rotation`, and IAM auth toggle all show real care.
- **`aws/modules/transit-gateway`** is clean and correctly uses `configuration_aliases = [aws.owner, aws.accepter]`.
- **Procedure scripts** (`failover.sh`, `failback.sh`, `verify_dual_region.sh`, `export_environment_prerequisites.sh`) consistently use `set -euo pipefail`, `: "${VAR:?...}"` guards, and structured `log()`/`err()` helpers — high quality.
- **`postgres_seed.tf`** seeding is idempotent (`pg_roles` / `pg_auth_members` existence checks, `GRANT rds_iam` guarded) and the `db_seed_run_id` trigger for re-runs is a nice touch.
- **`data.aws_region.*.id`** (rather than `.name`) is the correct, non-deprecated form for AWS provider v6 — good catch on the deprecation fixes.
- **README** clearly marks the arch as experimental / not-for-production and documents the known limitations (no Identity/Keycloak/Optimize). 👍
- The configurable `container_health_check_start_period_seconds` for slow Raft quorum formation is the right knob to expose.

---

### ❓ Questions / smaller items

- **Image bump `8.9.0-alpha5` → `8.9.0`** in the shared module default: I confirmed the tag exists (multi-arch). Just note it also changes the default for **single-region** — intended as part of release-duty, or should the dual-region ref pin its own image and leave the shared default until release time?
- **`enable_internal_nlb_raft_listener` + `internal_nlb_arn`**: the listener uses `raft_26502[0]` guarded by the bool — good. Worth a `precondition` that `internal_nlb_arn != ""` when the listener is enabled.
- **No CI ran on this PR** (`skip_all` is checked in the workflow checklist). Before it leaves draft, it'd be good to let `terraform validate`/golden/tflint/trivy run — trivy in particular will have opinions about the S3 policy in (1), and there are already a couple of `#trivy:ignore` pragmas in `s3.tf` that should be justified.
- **Relationship to #2271**: both add Transit Gateway support. This PR creates a *managed* TGW via a reusable module; #2271 attaches to a *pre-existing* TGW inline. Aligning on one TGW approach (ideally the shared module) would avoid two divergent patterns across the AWS refs.

---

Given the scope, keeping this as a draft is the right call. The big-ticket items before it's review-ready: **(1)** fix the shared-module S3 policy so single-region isn't weakened, **(2)** collapse the `clusters/` vs `infra/`+`app/` duplication, **(3)** pull the planning docs / reconcile `CLAUDE.md`, **(4)** resolve the merge conflicts, and **(5)** let CI run. Happy to do a deep re-review on the Terraform once the layout is consolidated.

aws/modules/ecs/fargate/orchestration-cluster/s3.tf
```diff
@@ -71,14 +66,6 @@ resource "aws_iam_policy" "s3_access" {
           "s3:DeleteObject"
         ]
         Resource = "${aws_s3_bucket.main.arn}/*"
-        Condition = {
-          StringEquals = {
-            "aws:RequestedRegion" = var.aws_region
-          }
-          Bool = {
-            "aws:SecureTransport" = "true"
```
leiicamundi, line 79: 🔴 **Security regression.** Removing `aws:SecureTransport` drops TLS-only enforcement on S3 object access, and this is the **shared** module also used by `ecs-single-region-fargate` — so single-region is weakened too. The `aws:RequestedRegion` / `kms:ViaService` 403s come from the cross-region access pattern; please **re-scope** these conditions (allow both regions) instead of removing them, and keep `SecureTransport`. See the summary comment for details.

CLAUDE.md
```diff
@@ -0,0 +1,361 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
```
leiicamundi, line 3: On `main`, `CLAUDE.md` is intentionally a 7-line stub that delegates to `AGENTS.md`. Replacing it with a 361-line standalone file breaks that convention and is a merge-conflict source. Suggest folding genuinely new content into `AGENTS.md` and keeping the stub.

docs/plans/2026-04-09-001-feat-ecs-dual-region-rdbms-plan.md
```diff
@@ -0,0 +1,552 @@
+---
```
leiicamundi: These internal planning/brainstorm docs (`docs/{plans,brainstorms,superpowers}`) aren't on `main` and aren't gitignored — they look like working artifacts rather than reference documentation. Recommend removing them from the PR or gitignoring the directories.

aws/containers/ecs-dual-region-fargate/terraform/infra/vpc.tf
```diff
@@ -0,0 +1,53 @@
+################################
```
leiicamundi: `terraform/infra/` + `terraform/app/` duplicate `terraform/clusters/` (13 `.tf` files are byte-identical; `clusters/` = `infra/` + `camunda.tf` + `iam.tf`). The README and all `.claude` commands only drive `clusters/`, so this split is currently unreferenced. Recommend keeping a single layout.

aws/modules/ecs/fargate/orchestration-cluster/variables.tf
```diff
@@ -195,7 +201,7 @@ variable "image" {
   # TODO: [release-duty] before the release, update the below versions to the stable release!
   # TODO: [release-duty] adjust renovate comment to bump the minor version to the new stable release
   # renovate: datasource=docker depName=camunda/camunda versioning=regex:^8\.9\.(?<patch>\d+)(-alpha(?<prerelease>[1-9]|[1-4][0-9]|50))?$
-  default = "camunda/camunda:8.9.0-alpha5"
+  default = "camunda/camunda:8.9.0"
```
leiicamundi: Confirmed `camunda/camunda:8.9.0` exists (multi-arch) — good. Note this default change also affects the **single-region** ref-arch that shares this module: intended as release-duty, or should the dual-region ref pin its own image until release time?

leiicamundi (commented):

leiicamundi:
### Missing MAP `map-migrated` tag on KMS and ELB resources

This PR adds a new ECS dual-region Fargate architecture that creates **KMS keys** and **Elastic Load Balancers** (ALB/NLB), none of which receive the AWS MAP tag `map-migrated = "migARUADZHVWZ"`. For example:

- KMS: `aws_kms_key.secrets_region_{0,1}`, `aws_kms_key.s3_region_{0,1}` (+ aliases)
- ELB: `aws_lb.alb_region_{0,1}`, `aws_lb.nlb_grpc_region_{0,1}`, `aws_lb.nlb_raft_region_{0,1}`

**Why it matters** — Camunda is enrolled in the AWS Migration Acceleration Program (MAP), which credits a share of our eligible AWS spend, but only for resources tagged `map-migrated = migARUADZHVWZ`. KMS and Elastic Load Balancing are eligible services, so untagged keys and load balancers are excluded from the credit. Context + FinOps Slack thread: camunda/team-infrastructure-experience#1151.

**How it's done elsewhere** — the tag is injected into the provider `default_tags` by CI, so it propagates to every taggable resource (KMS, LB, …). The providers added here already wire `default_tags = var.default_tags`; only the value is missing:

- Single-region EKS / ECS-Fargate add it to the `tags:` input:
  - `.github/workflows/aws_kubernetes_eks_single_region_tests.yml:224`
  - `.github/workflows/aws_ecs_single_region_fargate_tests.yml:162`
- EKS dual-region sets it directly as a Terraform var:
  - `.github/workflows/aws_kubernetes_eks_dual_region_tests.yml:73` → `TF_VAR_default_tags: '{"map-migrated": "migARUADZHVWZ"}'`

**How it should be done here** — this architecture's Terratest harness builds `default_tags` from the Go `commonTags` map (currently only `Test` / `RunID`). Terratest passes these as `-var`, which overrides any `TF_VAR_default_tags` env var, so the tag has to be added to `commonTags`:

```go
commonTags := map[string]interface{}{
    "Test":         "true",
    "RunID":        clusterPrefix,
    "map-migrated": "migARUADZHVWZ",
}
```

in each of:
- `aws/containers/ecs-dual-region-fargate/test/src/dual_region_greenfield_opensearch_test.go`
- `aws/containers/ecs-dual-region-fargate/test/src/dual_region_greenfield_rdbms_test.go`
- `aws/containers/ecs-dual-region-fargate/test/src/byo_vpc_test.go`
- `aws/containers/ecs-dual-region-fargate/test/src/failback_test.go`
- `aws/containers/ecs-dual-region-fargate/test/src/failover_test.go`

`commonTags` is already forwarded into `helpers.SetupBYOVPCs`, so the BYO-VPC path is covered too. With that in place the new KMS keys and load balancers inherit the MAP tag like every other AWS architecture.

Ref: camunda/team-infrastructure-experience#1151

hisImminence (approved):
Lets go! Nice work and looking forward for feedback from first internal users (e.g. Dave)

PR #2349, opened 2026-04-21
fix: add retry logic for rosa delete operator-roles in ghost cluster …

leiicamundi:
## TODO
- [ ] backport 8.8, 8.9 and main

Related to https://camunda.slack.com/archives/C076N4G1162/p1776748795476399

After 'rosa delete cluster --watch' returns, the ROSA backend may still consider the cluster as using the operator roles for a short period. This causes 'rosa delete operator-roles' to fail immediately with:
  ERR: There are clusters using Operator Roles Prefix '...', can't delete the IAM roles

Add a retry loop (up to 10 attempts, 60s apart) to handle this race condition.


leiicamundi:
We finally have a working cleanup https://github.com/camunda/camunda-deployment-references/actions/runs/24782167032/job/72516167629?pr=2349 - I'll have to backport it

PR #2417, opened 2026-05-06
fix(aws-ecs): unblock Fargate scenario — valid image pin, deprecation fix, fail-fast, debug action

marcel-dias:
Unblocks the AWS Containers ECS Single Region Fargate scenario, prompted by the failure of [run 25416825553](https://github.com/camunda/camunda-deployment-references/actions/runs/25416825553) and follow-up runs that retried the failing task 18+ times.

**Root cause and fixes**:
- **`fix(aws-ecs)` — pin a tag that exists**: `camunda/camunda:8.10.0` and `camunda/connectors-bundle:8.10.0-SNAPSHOT` are not published on Docker Hub yet (8.10.x is at `8.10.0-alpha1-rc2` / rolling `8.10-SNAPSHOT`). Pin both module defaults to the rolling `8.10-SNAPSHOT` tag — published, tracking, and replaceable when 8.10 GA ships (the existing `[release-duty]` TODO comments capture the cleanup).
- **`fix(aws-ecs)` — fail fast on broken deployments**: enable `deployment_circuit_breaker { enable = true, rollback = false }` on both `aws_ecs_service` resources (orchestration-cluster + connectors), and lower `service_timeouts.create` from `30m` → `15m`. ECS now marks the deployment FAILED at AWS' computed threshold (~10 failed task starts) instead of grinding the full 30 minutes.
- **`fix(aws)` — silence and pre-empt v7 break**: replace deprecated `data.aws_region.current.name` → `.region` in the ECS scenario's `s3.tf` and across all 11 occurrences in the shared `aws/common/cognito/outputs.tf` (no-op rename under `hashicorp/aws ~> 6.0`).
- **`feat(ci)` — visibility on next failure**: new composite action `internal-debug-failed-ecs-tasks` mirrors `internal-debug-failed-pods`. On `prepare-clusters` failure it dumps cluster description, per-service deployments + last 30 events (the most informative source for stabilization-timeout root-causing), RUNNING/STOPPED tasks with `stoppedReason`/exit codes, task definitions, and CloudWatch log streams under `/ecs/<prefix>` — tail in CI output, full content as `ecs-debug-dump-<distro>-<scenario>` artifact.
- **`chore(repo)`**: gitignore `.claude/` local agent state.

Golden plan refreshed to reflect the new `deployment_circuit_breaker` block on both ECS services. The image pin and `service_timeouts.create` change are not visible in the golden plan (`container_definitions` is computed; `timeouts` is a wait-config meta-argument).


hisImminence (changes requested):
Two things:

1. max-stopped-tasks — description vs. implementation mismatch

The input is documented as "Maximum number of recently-stopped tasks to inspect per service", but the implementation runs list-tasks at the cluster level:

TASK_ARNS=$(aws ecs list-tasks --cluster "$CLUSTER" \
  --desired-status "$status" --max-items "$MAX_STOPPED" \  # ← cluster-wide, not per-service
  --query 'taskArns' --output text || true)


Either the description should say "per cluster", or the --max-items cap should move inside the per-service loop. Small thing but could be confusing when debugging a multi-service cluster.

2. Debug action only wired on prepare-clusters failure

The action fires when prepare-clusters fails — which covers the stabilization timeout that triggered this PR. But a service can also fail to reach STABLE during the integration-tests job (e.g. tasks crashing under load, health check failures after smoke tests start hitting the ALB). In that case we'd still get no ECS debug output.

Should we wire the same action into integration-tests on failure too? Would make the next silent failure a lot easier to root-cause.

marcel-dias:
> Two things:
> 
> 1. max-stopped-tasks — description vs. implementation mismatch
> 
> The input is documented as "Maximum number of recently-stopped tasks to inspect per service", but the implementation runs list-tasks at the cluster level:
> 
> TASK_ARNS=$(aws ecs list-tasks --cluster "$CLUSTER" --desired-status "$status" --max-items "$MAX_STOPPED" \ # ← cluster-wide, not per-service --query 'taskArns' --output text || true)
> 
> Either the description should say "per cluster", or the --max-items cap should move inside the per-service loop. Small thing but could be confusing when debugging a multi-service cluster.
> 
> 2. Debug action only wired on prepare-clusters failure
> 
> The action fires when prepare-clusters fails — which covers the stabilization timeout that triggered this PR. But a service can also fail to reach STABLE during the integration-tests job (e.g. tasks crashing under load, health check failures after smoke tests start hitting the ALB). In that case we'd still get no ECS debug output.
> 
> Should we wire the same action into integration-tests on failure too? Would make the next silent failure a lot easier to root-cause.

Modified to address your comments

hisImminence (approved):

PR #2418, opened 2026-05-06
fix(helm): rename keys removed in camunda-platform chart 15.0.0

marcel-dias:
## Summary

The Wednesday scheduled run of `local_kubernetes_kind_single_region_tests` ([run 25417637176](https://github.com/camunda/camunda-deployment-references/actions/runs/25417637176)) went red on all four matrix scenarios because chart 15.0.0 (Camunda 8.10) removes two keys we still set:

- `global.ingress.host` → renamed to `global.host` (sibling of `global.ingress`)
- `webModeler.restapi.externalDatabase.user` → renamed to `username`

The chart's `templates/common/constraints.tpl` does `hasKey` checks and calls `fail`, so `helm upgrade --install` aborts before the cluster is even touched.

## Verification

- `helm template` against `oci://registry.camunda.cloud/team-distribution/camunda-platform:15-dev-latest` for all four matrix combinations (`domain`, `no-domain`, `pg-only-domain`, `pg-only-no-domain`) — exit 0, zero `[camunda][error|warning]` lines, rendered Ingress shows `host: camunda.example.com`.
- This PR only touches the kind single-region values; the same renames are needed in `aws/`, `azure/`, `tests/utils/`, and migration tests, but those feed different workflows and will be split into per-architecture follow-up PRs.

## Test plan

- [ ] CI: all four `Test Kind - *` jobs go green
- [ ] CI: existing `Check for Helm deprecation warnings` step passes (no other deprecations surface)

### DIFF
--- a/.github/workflows/local_kubernetes_kind_single_region_tests.yml
+++ b/.github/workflows/local_kubernetes_kind_single_region_tests.yml
@@ -238,6 +238,14 @@ jobs:
               with:
                   release-name: ${{ env.CAMUNDA_RELEASE_NAME }}
                   namespace: ${{ env.CAMUNDA_NAMESPACE }}
+                  # The Camunda Hub consolidation (chart 15.x) deprecates the legacy
+                  # console.enabled / webModeler.enabled flags. Migrating to
+                  # camundaHub.enabled changes deployment topology and ingress paths,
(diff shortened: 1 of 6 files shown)


hisImminence (approved):
✅

PR #2450, opened 2026-05-13
fix(ci): don't fail helm deprecation check on non-PR events

marcel-dias:
## Summary

Extends the scope of the `internal-helm-deprecation-check` action so it is **non-blocking on every event**, and adds a **Slack alert on (simulated) scheduled runs**.

Folds in #2513 (closed in favor of this PR).

### Behavior

The action now reports findings in an event-aware, never-failing way:

- **pull_request / pull_request_target / workflow_run** — posted to a single shared PR comment (one section per workflow/job/release/namespace combination). _(unchanged from #2419.)_
- **schedule (real or simulated via a `schedules/*` head_ref)** — emitted as `::warning::` annotations and, when the caller wires `vault-*` + `slack-channel-id`, posted to Slack via `report-failure-on-slack`.
- **workflow_dispatch / push / any other event** — emitted as `::warning::` annotations in the job log only.

The Slack-alert step is gated on the caller workflow's **`IS_SCHEDULE`** env var rather than `github.event_name == 'schedule'`. This is the convention already used elsewhere in the repo:

```yaml
env:
  IS_SCHEDULE: ${{ (contains(github.head_ref, 'schedules/') || github.event_name == 'schedule') && 'true' || 'false' }}
```

That way, our `[schedule][NO MERGE]` simulation PRs (head_ref `schedules/*`) also trigger the end-to-end Slack path during verification, instead of being silently skipped.

### Why

Per the Wednesday 2026-05-20 schedule snapshot, the strict unknown-keys check on chart `camunda-platform-14.2.0` accounts for ~24 of 27 failed jobs even when the cluster is healthy — purely a CI/visibility regression. Making the check non-blocking turns these into actionable warnings (logged + Slacked on schedule) without holding up release pipelines.

## Test plan

- [ ] Trigger via a `schedules/*` PR and confirm the Slack alert fires when findings are present
- [ ] Trigger via `workflow_dispatch` and confirm warnings appear in the log but no Slack post
- [ ] Trigger via a regular PR and confirm the shared PR-comment path still works
- [ ] Confirm the workflow exits 0 in all three cases when findings are present


leiicamundi:
Workflows are now updated to forward `VAULT_ADDR` / `VAULT_ROLE_ID` / `VAULT_SECRET_ID` from `secrets.*` into the composite action (and through `internal-multi-region-tests` for dual-region callers). Composite actions can't read `secrets.*` directly, so this wiring is required for scheduled runs to actually fire Slack alerts via `report-failure-on-slack`. Slack channel defaults remain hardcoded inside the action (`infraex-alerts` on schedule, `infraex-test` otherwise), so workflows only need to pass the vault inputs.

PR #2505, opened 2026-05-22
fix(generic-debian): vendor connectors launcher to avoid Spring Boot jar mismatch

leiicamundi:
## Summary

- Drop the `curl … connectors@main/start.sh` + `sed` rewrite in `generic/compute/debian/procedure/camunda-install.sh`.
- Vendor a minimal launcher that does `exec java \${JAVA_OPTS:-} -jar /opt/camunda/connectors/connectors.jar`.
- In `generic/compute/debian/procedure/camunda-checks.sh`, dump `systemctl status` + `journalctl` when a service is reported as not running, so the next regression is self-diagnosing.

## Background

The connector-runtime-bundle is a Spring Boot uber-jar with dependencies under `BOOT-INF/lib/`. The previous install script fetched `start.sh` from `camunda/connectors@main` unpinned and then `sed`-replaced its last line with a flat-classpath invocation:

```bash
java ${JAVA_OPTS} -cp "/opt/camunda/connectors/*" \
  io.camunda.connector.runtime.app.ConnectorRuntimeApplication
```

Flat classpath cannot resolve classes inside `BOOT-INF/lib/`, so the JVM dies on launch, systemd loops on `Restart=on-failure`, and `127.0.0.1:9090/actuator/health` never opens. Earlier connectors releases happened to ship a flat tarball, masking the bug.

**Smoking gun in run [26194101647](https://github.com/camunda/camunda-deployment-references/actions/runs/26194101647)**: `TestCamundaUpgrade` installs connectors `8.8.0` → service starts; then upgrades to `8.9.4` (introduced in `0cf989cc`) → same procedure, same EC2 host, but now `[FAIL] camunda-connectors.service is not running`. Proof that only the bundle layout regressed; the procedure is the wrong shape.

## Why hardcode `/opt/camunda/connectors/connectors.jar`?

`MNT_DIR` defaults to `/opt/camunda` in `camunda-install.sh:21`, and the systemd unit `generic/compute/debian/configs/camunda-connectors.service:11` already hardcodes `ExecStart=/opt/camunda/connectors/start.sh`. Anyone overriding `MNT_DIR` would already have to patch the unit file; the launcher just matches.

## Test plan

- [ ] Run `aws_compute_ec2_single_region_tests.yml` (x86_64 + arm64) and verify `camunda-connectors.service` reaches `active (running)` and `127.0.0.1:9090/actuator/health` returns HTTP 200.
- [ ] Run `TestCamundaUpgrade` and confirm the 8.8.0 → 8.9.4 upgrade path succeeds.
- [ ] If a service still fails, confirm the new diagnostic block prints `journalctl` lines in the workflow log.

## Backport

Required on `stable/8.9` and `stable/8.8` (same RA). Will be included in a bundled backport PR after this lands.

### DIFF
--- a/aws/compute/ec2-single-region/test/src/default_ec2_test.go
+++ b/aws/compute/ec2-single-region/test/src/default_ec2_test.go
@@ -344,11 +344,17 @@ func TestCamundaUpgrade(t *testing.T) {
 		t.Fatalf("Error writing file: %v", err)
 	}
 
-	// Zeebe has a prerelease protection that results in unhealthy clusters if not disabled
+	// Both Zeebe broker and the search-engine SchemaManager refuse upgrades from/to a
+	// pre-release version (see io.camunda.zeebe.util.migration.VersionCompatibilityCheck —
+	// UseOfPreReleaseVersion). Disable both gates so the upgrade test can run against
+	// SNAPSHOT/alpha builds.
 	if strings.Contains(camundaCurrentVersionFull, "SNAPSHOT") || strings.Contains(camundaCurrentVersionFull, "alpha") {
 		cmd = shell.Command{
 			Command: "bash",
-			Args:    []string{"-c", "echo ZEEBE_BROKER_EXPERIMENTAL_VERSIONCHECKRESTRICTIONENABLED=false >> ../../configs/camunda-environment"},
+			Args: []string{"-c", `cat >> ../../configs/camunda-environment <<'EOF'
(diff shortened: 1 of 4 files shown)


leiicamundi:
Added a code comment on `TestCamundaUpgrade` documenting that broker startup failures during the in-place upgrade are typically caused by broken upstream SNAPSHOT artifacts at https://artifacts.camunda.com/artifactory/zeebe/io/camunda/camunda-zeebe/ — not something fixable in this repo. If this is the only red check on an otherwise green run, re-run later or check upstream Zeebe CI.

PR #2510, opened 2026-05-22
backport(stable/8.9): vendor EC2 connectors launcher + schema-manager pre-release bypass

leiicamundi:
## Summary

Backport of [#2505](https://github.com/camunda/camunda-deployment-references/pull/2505) to `stable/8.9`. Two cherry-picks:

- `ac9129ec` — vendor a minimal connectors `start.sh` (`exec java -jar connectors.jar`) instead of downloading the unpinned upstream `start.sh` and `sed`-rewriting it into a flat `java -cp ...` launcher. The flat-classpath approach broke from connectors `8.9.4` onward because the bundle ships as a Spring Boot uber-jar (`BOOT-INF/lib/`), causing `camunda-connectors.service` to fail on every fresh install (root cause of the recurring EC2 scheduled run failures — see `./debug/ci-failures-report.md`).
- `665120cf` — add `CAMUNDA_DATABASE_SCHEMA_MANAGER_VERSION_CHECK_RESTRICTION_ENABLED=false` alongside the existing Zeebe broker flag in `TestCamundaUpgrade`'s SNAPSHOT/alpha branch. No-op on `stable/8.9` (pinned to `8.9.5`), kept for symmetry with `main`.

Also adds richer service-down diagnostics to `camunda-checks.sh` (`systemctl status` + `journalctl` dump).

## Docs

Verified: no docs in `generic/compute/debian/**`, `aws/compute/ec2-single-region/**` or repo-root `docs/` reference `start.sh`, `JAVA_OPTS`, `ConnectorRuntimeApplication` or the upstream `camunda/connectors` repo. No internal doc change needed.

External `camunda/camunda-docs` follow-up: not required for this fix — the user-facing procedure is unchanged.

## Test plan

- [ ] Next scheduled EC2 Single Region run on `stable/8.9` passes `TestCamundaUpgrade`, `TestAllInOneScript`, `TestCloudWatchFeature` on both x86_64 and arm64
- [ ] `camunda-connectors.service` reaches `Active: running` on fresh install with `CAMUNDA_CONNECTORS_VERSION=8.9.4`

### DIFF
--- a/aws/compute/ec2-single-region/test/src/default_ec2_test.go
+++ b/aws/compute/ec2-single-region/test/src/default_ec2_test.go
@@ -344,11 +344,17 @@ func TestCamundaUpgrade(t *testing.T) {
 		t.Fatalf("Error writing file: %v", err)
 	}
 
-	// Zeebe has a prerelease protection that results in unhealthy clusters if not disabled
+	// Both Zeebe broker and the search-engine SchemaManager refuse upgrades from/to a
+	// pre-release version (see io.camunda.zeebe.util.migration.VersionCompatibilityCheck —
+	// UseOfPreReleaseVersion). Disable both gates so the upgrade test can run against
+	// SNAPSHOT/alpha builds.
 	if strings.Contains(camundaCurrentVersionFull, "SNAPSHOT") || strings.Contains(camundaCurrentVersionFull, "alpha") {
 		cmd = shell.Command{
 			Command: "bash",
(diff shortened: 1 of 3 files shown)


leiicamundi:
Added a code comment on `TestCamundaUpgrade` documenting that broker startup failures during the in-place upgrade are typically caused by broken upstream SNAPSHOT artifacts at https://artifacts.camunda.com/artifactory/zeebe/io/camunda/camunda-zeebe/ — not something fixable in this repo. If this is the only red check on an otherwise green run, re-run later or check upstream Zeebe CI.

PR #2513, opened 2026-05-22
feat(ci): make internal-helm-deprecation-check non-blocking, alert Slack on schedule

leiicamundi:
## Summary

Rework the `internal-helm-deprecation-check` composite action so it never fails the workflow. Findings are now reported event-aware:

- **pull_request / workflow_dispatch / push**: emit `::warning::` annotations in the job log only (no PR comment, no failure).
- **schedule**: same `::warning::` annotations, plus a Slack alert via the shared `report-failure-on-slack@1.7.0` action, gated on the caller wiring the new optional inputs (`vault-addr`, `vault-role-id`, `vault-secret-id`, `slack-channel-id`, `slack-mention-people`).

Goal: trace deprecation/unknown-key drift in the deployed chart without blocking release pipelines.

Supersedes #2450. Refs https://github.com/camunda/camunda-platform-helm/issues/4564.

### Backports

- `stable/8.9`: separate PR opened (this rework was hand-crafted against the 8.9 variant which never had #2419's PR-comment infrastructure).
- `stable/8.8` and `stable/8.7`: action is **not** present on those branches — no backport required.

## Test plan

- [ ] Trigger a scheduled run with the new inputs wired and verify Slack alert fires when the deployed chart contains a `[camunda][warning]` note or an unknown key.
- [ ] Trigger a PR run and confirm the workflow stays green even with findings (warning annotations visible in the job log).
- [ ] Trigger a `workflow_dispatch` and confirm same behavior as PR.

### DIFF
--- a/.github/actions/internal-helm-deprecation-check/README.md
+++ b/.github/actions/internal-helm-deprecation-check/README.md
@@ -9,10 +9,17 @@ Uses `helm get notes` to retrieve the release notes and scans for
 Optionally validates that deployed values do not contain unknown keys
 by checking against a strict version of the chart's JSON Schema.
 
-On pull request events, findings are reported as a single shared PR
-comment (one comment per PR, with one section per
-workflow/job/release/namespace combination) and the workflow is NOT
-failed. On non-PR events, findings still fail the workflow.
+Findings are reported in an event-aware, non-blocking way:
+  - pull_request / pull_request_target / workflow_run: posted to a
+    single shared PR comment (one comment per PR, one section per
+    workflow/job/release/namespace combination).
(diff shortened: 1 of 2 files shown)


leiicamundi:
Folding scope into #2450 — that PR now covers both the warn-only-on-non-PR-events change (its original scope) and this PR's non-blocking-everywhere + Slack-on-schedule rework. Also switched the Slack-alert gate from `github.event_name == 'schedule'` to the workflow-level `IS_SCHEDULE` env var, so simulated schedules (head_ref `schedules/*`) also trigger the Slack path end-to-end during verification. Closing in favor of #2450.

PR #2514, opened 2026-05-22
fix(ci): make internal-helm-deprecation-check non-blocking on stable/8.9, alert Slack on schedule

leiicamundi:
## Summary

Backport of #2513 to `stable/8.9`. Rework `internal-helm-deprecation-check` so it never fails the workflow:

- **pull_request / pull_request_target / workflow_run**: posted to a single shared PR comment (one section per workflow/job/release/namespace combination).
- **schedule**: `::warning::` annotations, plus a Slack alert via `report-failure-on-slack@1.7.0`, gated on the caller wiring the new optional inputs (`vault-addr`, `vault-role-id`, `vault-secret-id`, `slack-channel-id`, `slack-mention-people`).
- **workflow_dispatch / push / any other event**: `::warning::` annotations in the job log only.

The PR-comment infrastructure from #2419 (previously absent on `stable/8.9`) is now ported over so PRs against `stable/8.9` get the same shared comment as on main.

`stable/8.8` and `stable/8.7` do not have this action — no further backports required.

Refs https://github.com/camunda/camunda-platform-helm/issues/4564.

## Test plan

- [ ] Trigger a scheduled run on `stable/8.9` and verify Slack alert fires when findings are present.
- [ ] Trigger a PR run and confirm the workflow stays green even with findings, and the shared PR comment is posted/updated.


leiicamundi:
Workflows are now updated to forward `VAULT_ADDR` / `VAULT_ROLE_ID` / `VAULT_SECRET_ID` from `secrets.*` into the composite action (and through `internal-multi-region-tests` for dual-region callers). Composite actions can't read `secrets.*` directly, so this wiring is required for scheduled runs to actually fire Slack alerts via `report-failure-on-slack`. Slack channel defaults remain hardcoded inside the action (`infraex-alerts` on schedule, `infraex-test` otherwise), so workflows only need to pass the vault inputs.

leiicamundi:
Backported the PR-comment infrastructure from #2419 (`upsert-pr-comment.py` + the `💬 Report findings` step + `comment-section-key` / `github-token` inputs) — pushed as `cbe1f005`. The next CI run against this branch should now produce a shared findings comment on the PR when deprecations/unknown-keys are detected.

PR #2525, opened 2026-05-22
fix(test): drop demo:demo default in check-zeebe-cluster-topology.sh (INC-5340)

leiicamundi:
## Summary

Backport of the topology-script fix landed on main (#2523) and stable/8.9 (#2524).

`generic/kubernetes/dual-region/procedure/check-zeebe-cluster-topology.sh` unconditionally sent \`-u demo:demo\` when \`ZEEBE_BASIC_AUTH_USER/PASSWORD\` were unset. Replaced with a conditional \`-u\` so:

- an unauthenticated cluster (current CI mode on \`rosa-hcp-dual-region\`) works without auth header,
- an authenticated cluster receives only real credentials from the caller.

Note: \`http-smoke-test.sh\` does not exist on stable/8.8, so only the topology script is touched here.

## Test plan

- [x] \`bash -n\` passes on the modified script.
- [x] Pre-commit hooks pass.
- [ ] \`aws_openshift_rosa_hcp_dual_region_tests.yml\` re-runs green.

### DIFF
--- a/generic/kubernetes/dual-region/procedure/check-zeebe-cluster-topology.sh
+++ b/generic/kubernetes/dual-region/procedure/check-zeebe-cluster-topology.sh
@@ -15,7 +15,14 @@ trap cleanup EXIT
 sleep 2
 
 echo "📡 Fetching Zeebe cluster topology..."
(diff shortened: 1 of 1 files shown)


leiicamundi:
Closing: the topology script is a customer-facing procedure and we don't want to change its documented behaviour. The matching follow-ups on main (#2526) and stable/8.9 (#2527) have been re-scoped to the smoke-test only.

PR #2574, opened 2026-06-01
fix(eks-dual-region): evaluate basic-auth header lazily (INC-5340)

leiicamundi:
## Description

The INC-5340 hardening introduced a package-level guard for the Camunda
basic-auth credentials in the EKS dual-region test helpers:

```go
var basicAuthDemoHeader = func() string {
    user := os.Getenv("CAMUNDA_BASIC_AUTH_USER")
    password := os.Getenv("CAMUNDA_BASIC_AUTH_PASSWORD")
    if user == "" || password == "" {
        panic("CAMUNDA_BASIC_AUTH_USER and CAMUNDA_BASIC_AUTH_PASSWORD must be set (see INC-5340)")
    }
    ...
}()
```

Because it is evaluated **at package init**, *any* `go test` invocation that
imports the `kubectl` helper package panics when the credentials are absent —
including phases that never issue an authenticated request.

In particular, the `prepare-clusters` and `cleanup-clusters` jobs only run
`go test -run TestAWSKubeConfigCreation` (kubeconfig generation) and only fetch
`CI_ENCRYPTION_KEY` from Vault — never the Camunda credentials. As a result both
jobs fail at init with:

```
panic: CAMUNDA_BASIC_AUTH_USER and CAMUNDA_BASIC_AUTH_PASSWORD must be set (see INC-5340)
```

## Fix

Convert `basicAuthDemoHeader` from an init-evaluated variable into a function
that is evaluated **on demand** at each authenticated request. The guard now
only trips when an authenticated request is actually made, so kubeconfig
generation and cleanup no longer require the credentials.

The INC-5340 protection (the publicly reachable CI cluster must never
authenticate with `demo:demo`) is fully preserved for every real request.

## Changes

- `basicAuthDemoHeader` → `basicAuthDemoHeader()` (lazy function).
- Updated all 9 call sites to invoke the function.

## Validation

- `go build ./...` and `go vet ./internal/helpers/kubectl/` pass.
- No remaining bare references to the variable.

## Backports

- `stable/8.9`: #2573 (variable named `basicAuthHeader`).
- `stable/8.8` and `stable/8.7`: not affected — the `eks-dual-region` test
  helper does not exist on those branches.

### DIFF
--- a/aws/kubernetes/eks-dual-region/test/internal/helpers/kubectl/helpers.go
+++ b/aws/kubernetes/eks-dual-region/test/internal/helpers/kubectl/helpers.go
@@ -27,19 +27,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// basicAuthDemoHeader is the Authorization header for the Camunda basic-auth user
-// used by the test suite. Sourced from CAMUNDA_BASIC_AUTH_USER and
-// CAMUNDA_BASIC_AUTH_PASSWORD env vars (set in CI by internal-camunda-ci-credentials).
-// Panics if either is empty — the publicly reachable CI cluster must never
-// authenticate with demo:demo (see incident INC-5340).
-var basicAuthDemoHeader = func() string {
+// basicAuthDemoHeader returns the Authorization header for the Camunda basic-auth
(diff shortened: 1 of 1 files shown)


leiicamundi:
Merging as it's only tests related

PR #2578, opened 2026-06-02
fix(eks-dual-region): drop removed identityKeycloak values key (chart 15)

leiicamundi:
## Description

The EKS dual-region integration test (`TestAWSDeployDualRegCamunda/TestDeployC8Helm`)
fails when deploying the Camunda Platform Helm chart `15-dev-latest`
(chart 15.x / Camunda 8.10):

```
Error: execution error at (camunda-platform/templates/common/constraints.tpl):
[camunda][error] The Helm values file key "identityKeycloak" has been removed.
```

In chart 15.x the bundled Bitnami subcharts (`identityKeycloak`,
`identityPostgresql`, `webModelerPostgresql`, `elasticsearch`) were removed.
The constraint now fails on the mere **presence** of the `identityKeycloak`
key (`hasKey .Values "identityKeycloak"`), even when set to `enabled: false`.

## Fix

Remove the now-invalid top-level `identityKeycloak: { enabled: false }` block
from the dual-region values file. There is no bundled Keycloak subchart to
disable anymore, so the key is no longer needed.

The deployment intent is unchanged: this dual-region setup uses basic-auth
(`demo/demo`) with Identity authentication disabled
(`global.security.authentication.method: basic`, `identity.enabled: false`),
so no Keycloak is deployed regardless.

## Scope

Only the EKS dual-region values file is touched, since that is the scenario
failing in CI against the `15-dev-latest` chart. Other values files that still
reference bundled subcharts target different scenarios/chart versions and are
out of scope here.

## Validation

- `grep` confirms no remaining removed-subchart keys
  (`identityKeycloak`, `identityPostgresql`, `webModelerPostgresql`,
  `elasticsearch`) in `aws/kubernetes/eks-dual-region/helm-values/`.

### DIFF
--- a/aws/kubernetes/eks-dual-region/helm-values/camunda-values.yml
+++ b/aws/kubernetes/eks-dual-region/helm-values/camunda-values.yml
@@ -37,9 +37,6 @@ global:
 identity:
     enabled: false
 
-identityKeycloak:
-    enabled: false
-
 console:
     enabled: false # by default, console is not enabled
 
--- a/generic/kubernetes/operator-based/elasticsearch/camunda-elastic-values.yml
+++ b/generic/kubernetes/operator-based/elasticsearch/camunda-elastic-values.yml
@@ -37,7 +37,3 @@ optimize:
                 secret:
                     existingSecret: elasticsearch-es-elastic-user
                     existingSecretKey: elastic
-
-# Disable bundled Elasticsearch sub-chart (using ECK-managed instance)
-elasticsearch:
-    enabled: false


hisImminence (approved):

marcel-dias (approved):

leiicamundi:
## Follow-up: independent CI failures split into dedicated fix PRs

Triaged the remaining red checks on this PR — they are **independent** of this PR's change and reproduce on other PRs / main:

| Failure | Root cause | Fix |
|---|---|---|
| AKS single-region — `values-enterprise.yaml` HTTP 404 | The file was removed upstream for chart 15 (Camunda 8.10); it only configured the now-removed bundled Bitnami images | #2586 — drop the obsolete download, start from an empty base |
| migration-from-bitnami — `constraints.tpl: key "identityKeycloak" has been removed` | Source install reused chart `15-dev-latest`, which removed the bundled Bitnami sub-charts the migration is supposed to start from | #2587 — pin the migration **source** to chart 14.4.0 (last Bitnami-bundled line); target cutover stays on chart 15 |
| OpenShift ROSA HCP — `Waiting for std compute nodes completion finished with error` | Transient AWS-side slowness in compute-node provisioning makes the rhcs `wait_for_std_compute_nodes_complete` time out during `terraform apply` | dedicated PR incoming — convergent apply retry |

None require a change to this PR; the OpenShift one just needs a re-run until the dedicated fix lands.

PR #2586, opened 2026-06-04
ci(aks): drop obsolete values-enterprise.yaml base

leiicamundi:
## Root cause

The AKS single-region workflow downloaded the per-chart `values-enterprise.yaml` as the base `values.yml`. That overlay only ever configured the **bundled Bitnami sub-charts** (enterprise image registries for PostgreSQL / Keycloak / Elasticsearch).

Chart 15 (Camunda 8.10) **removed those sub-charts**, so `values-enterprise.yaml` no longer exists upstream:

```
charts/camunda-platform-8.10/values-enterprise.yaml -> 404
charts/camunda-platform-8.9/values-enterprise.yaml  -> 200
```

Every 8.10 AKS run therefore failed in **setup** with `curl` exit 22 (HTTP 404), before any Helm deploy.

## Fix

Remove the `values-enterprise.yaml` download **entirely** and start from an empty base. The deployment is built purely from the reference-architecture overlays merged afterwards (operator-managed Elasticsearch / Keycloak, OIDC) — which is what AKS actually deploys on chart 15. The enterprise Bitnami overlay is obsolete for this architecture.

## Scope

Only the AKS single-region workflow downloaded this file. EKS single-region and OpenShift carry a TODO to align but do **not** download it today, so they are unaffected.

> Draft — a maintainer should review and approve before merge.

PR #2588, opened 2026-06-04
fix(rosa): retry convergent apply on transient compute-node timeout (main)

leiicamundi:
## Problem

The ROSA HCP integration workflows (single- and dual-region) intermittently fail during cluster creation with:

```
Waiting for std compute nodes completion finished with error
```

This happens **after** the cluster itself has already been created: the rhcs provider's `wait_for_std_compute_nodes_complete` post-create wait times out because AWS-side compute-node provisioning was slow. It is a transient infrastructure delay, not a configuration error — re-running the job almost always succeeds.

Observed e.g. on #2578 (≈70 min, no code relation to that PR).

## Strategy: convergent apply retry (not a blind re-run)

The cluster `terraform apply` step is wrapped in a bounded retry (max 3 attempts):

- On the **first** attempt, the previously-saved plan is applied.
- If apply fails **and** the error matches a known compute-node provisioning timeout pattern, the state has already partially advanced (the cluster exists), so we **re-plan** against the current state and **re-apply**. Because the cluster is in the Terraform state, this simply re-enters the post-create wait on the *existing* cluster — nothing is recreated.
- Any error that is **not** a recognised transient timeout fails fast on the first attempt, so genuine misconfigurations are never masked.

```
transient_pattern='std compute nodes|compute nodes completion|waiting for (std )?compute|context deadline exceeded|timeout while waiting'
```

### Why retry and not just raise the wait timeout?

The provider's wait is already generous; raising it further would lengthen every successful run and still not cover pathological AWS slowness. A convergent retry only pays the extra cost when a timeout actually occurs and is self-healing.

## Files

- `.github/actions/aws-openshift-rosa-hcp-single-region-create/action.yml` — `Terraform Apply - Cluster`
- `.github/actions/aws-openshift-rosa-hcp-dual-region-create/action.yml` — `Terraform Apply - Clusters`

## Validation

- `yamllint` / pre-commit pass.
- Behavioural validation requires a real ROSA HCP run (CI) — the retry path only triggers on an actual provisioning timeout.

> Draft — a maintainer should review the retry bound / pattern before merge.

### DIFF
--- a/.github/actions/aws-openshift-rosa-hcp-dual-region-create/action.yml
+++ b/.github/actions/aws-openshift-rosa-hcp-dual-region-create/action.yml
@@ -342,7 +342,7 @@ runs:
               # Convergent apply with retry for transient ROSA HCP compute-node
               # provisioning timeouts (see the shared helper for the rationale).
               # The helper manages `errexit` locally during its retries.
-              source "${{ github.workspace }}/.github/scripts/rosa-hcp-convergent-apply.sh"
+              source "${{ github.action_path }}/../../scripts/rosa-hcp-convergent-apply.sh"
               rosa_hcp_convergent_apply clusters.plan terraform.tfvars
 
         - name: Export Terraform Outputs - Clusters
--- a/.github/actions/aws-openshift-rosa-hcp-dual-region-create/action.yml
+++ b/.github/actions/aws-openshift-rosa-hcp-dual-region-create/action.yml
@@ -337,10 +337,11 @@ runs:
               RHCS_TOKEN: ${{ inputs.rh-token }}
           shell: bash
           run: |
-              set -uo pipefail
+              set -euo pipefail
 
               # Convergent apply with retry for transient ROSA HCP compute-node
               # provisioning timeouts (see the shared helper for the rationale).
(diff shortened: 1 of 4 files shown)


leiicamundi:
## 🔗 Merge order & dependencies

This PR adds a convergent-apply retry for transient ROSA HCP compute-node timeouts. It is part of a coordinated set sharing a common chart-15 root cause (the `15-dev-latest` chart removed `identityKeycloak` / `elasticsearch` / `webModelerPostgresql` value keys, so `constraints.tpl` fails any overlay still referencing them).

**Recommended merge order (main):**

1. **#2586** — chart-15 compatibility (foundational unblocker; approved).
2. **#2595** — Zeebe topology check robustness.
3. **#2588** ← this PR — ROSA HCP convergent-apply retry → backports **#2589** (8.8), **#2590** (8.7).
4. **#2587** — migration-from-bitnami source pinned to chart 14.

> The `rosa-hcp-dual-region` **integration/deploy** failure here is the chart-15 issue (fixed by #2586), distinct from the cluster-**create** timeout this PR addresses. Rebase on `main` after #2586 merges for a clean run. The retry path itself can only be exercised on a real transient timeout.

PR #2620, opened 2026-06-05
feat(migration): add KEYCLOAK_TARGET_MODE=external for migrating Bitnami Keycloak to an external instance

eamonnmoloney:
## Motivation

The Bitnami → external-infrastructure migration scripts (`generic/kubernetes/migration`) already support routing **PostgreSQL** and **Elasticsearch** to external (non-operator) targets via `PG_TARGET_MODE=external` / `ES_TARGET_MODE=external`. **Keycloak**, however, was always deployed as a Keycloak Operator `Keycloak` CR — there was no way to migrate a bundled Bitnami Keycloak onto a Keycloak that is managed, standalone, or deployed via a Helm chart.

This matters for two audiences:

- **Customers** who run (or want to run) Keycloak as a managed/standalone/Helm instance rather than via the Keycloak Operator.
- **CI**: the Camunda Helm test suites need to stand up **arbitrary versions** of Keycloak / PostgreSQL / Elasticsearch to mimic the spread of real customer environments — which operators don't make easy, but external (companion) instances do. Exercising these scripts in `external` mode is how we prove a Bitnami-8.9 customer can reach 8.10.

## What this PR does

Adds `KEYCLOAK_TARGET_MODE=external`:

- **`env.sh`** — `KEYCLOAK_TARGET_MODE` toggle + `EXTERNAL_KEYCLOAK_*` connection vars (protocol/host/port/context-path/realm/admin-secret).
- **`lib.sh`** — `is_external_keycloak`, `validate_external_keycloak_config`, and `generate_external_keycloak_values` (wires `global.identity.keycloak.url` at the external instance and sets `identityKeycloak.enabled: false`); `sync_keycloak_admin_credentials` skips the operator-only `keycloak-initial-admin` reconcile in external mode while still generating the admin/client-token override.
- **`1-deploy-targets.sh`** — skips Keycloak Operator deployment in external mode.
- **`3-cutover.sh`** — applies the external Keycloak helm values during the cutover Helm switch.
- **`4-validate.sh` / `5-cleanup-bitnami.sh`** — skip the operator `Keycloak` CR readiness check in external mode.
- **`README.md`** — documents the new flag, the `EXTERNAL_KEYCLOAK_*` variables, and when to use them.

The realm itself still migrates via the external Keycloak **PostgreSQL database** (`EXTERNAL_PG_KEYCLOAK_*`), so `KEYCLOAK_TARGET_MODE=external` requires `PG_TARGET_MODE=external`. The scripts do not deploy or manage the Keycloak instance — only the realm DB is migrated and Camunda is rewired to it.

## Status / validation

- `bash -n` clean on all changed scripts.
- **Not yet validated on a live cluster.** `shellcheck`/`pre-commit` were not available in my environment — please run `pre-commit run --all-files`.
- Follow-ups: a kind-based migration test case for `external` Keycloak (the existing matrix covers operator targets only), and a companion docs PR in [camunda-docs](https://github.com/camunda/camunda-docs) `migration-from-bitnami/` once this lands.

## Related

This unblocks Helm CI work in `camunda/camunda-platform-helm` to drive these scripts (install 8.9 + bundled Bitnami → migrate to companion external infra in `external` mode → upgrade to 8.10 → verify). That CI PR will reference this one.

### DIFF
--- a/generic/kubernetes/migration/2-backup.sh
+++ b/generic/kubernetes/migration/2-backup.sh
@@ -72,16 +72,22 @@ do_pg_backup() {
     save_state "${component^^}_BACKUP_TIMESTAMP" "$TIMESTAMP"
 }
 
+# The backup reads the SOURCE database, which may use different database/user
+# names than the migration target (e.g. the Camunda bundled Bitnami Keycloak
+# uses db "bitnami_keycloak"/user "bn_keycloak", while the target uses
+# "keycloak"/"keycloak"). The *_SOURCE_DB_NAME/_USER overrides decouple the
+# source names from the target names (which Phase 3 restore uses); they default
+# to the target names for the common same-name case.
 if [[ "${MIGRATE_IDENTITY}" == "true" ]]; then
-    do_pg_backup identity "${IDENTITY_DB_NAME}" "${IDENTITY_DB_USER}"
+    do_pg_backup identity "${IDENTITY_SOURCE_DB_NAME:-${IDENTITY_DB_NAME}}" "${IDENTITY_SOURCE_DB_USER:-${IDENTITY_DB_USER}}"
 fi
 
 if [[ "${MIGRATE_KEYCLOAK}" == "true" ]]; then
-    do_pg_backup keycloak "${KEYCLOAK_DB_NAME}" "${KEYCLOAK_DB_USER}"
+    do_pg_backup keycloak "${KEYCLOAK_SOURCE_DB_NAME:-${KEYCLOAK_DB_NAME}}" "${KEYCLOAK_SOURCE_DB_USER:-${KEYCLOAK_DB_USER}}"
 fi
 
 if [[ "${MIGRATE_WEBMODELER}" == "true" ]]; then
-    do_pg_backup webmodeler "${WEBMODELER_DB_NAME}" "${WEBMODELER_DB_USER}"
+    do_pg_backup webmodeler "${WEBMODELER_SOURCE_DB_NAME:-${WEBMODELER_DB_NAME}}" "${WEBMODELER_SOURCE_DB_USER:-${WEBMODELER_DB_USER}}"
 fi
 
 # ─────────────────────────────────────────────────────────────────────────────
--- a/generic/kubernetes/migration/3-cutover.sh
+++ b/generic/kubernetes/migration/3-cutover.sh
@@ -347,8 +347,17 @@ elif [[ "${SKIP_HELM_UPGRADE:-false}" == "true" ]]; then
     section "Data migration complete ($(timer_elapsed))"
     echo ""
     echo "Data has been migrated to the target backends."
-    echo "The Helm upgrade was skipped (SKIP_HELM_UPGRADE=true) — the caller is"
-    echo "responsible for upgrading Camunda to point at the new backends."
+    echo "Camunda is still frozen (scaled to 0) and pointed at the OLD backends —"
+    echo "the caller is responsible for the Helm upgrade that switches Camunda to"
+    echo "the new backends and restarts the components."
+    echo ""
+    echo "Phase 3 is marked complete so ./4-validate.sh and ./5-cleanup-bitnami.sh"
+    echo "can run — but only AFTER the caller's Helm upgrade has switched Camunda over."
+    # Mark phase 3 complete even though the Helm upgrade is deferred: the data
+    # migration that phase 3 owns IS done, so the phase state machine must not
+    # stay stuck (otherwise require_phase blocks the later phases entirely).
+    run_hooks "post-phase-3"
+    complete_phase 3
 else
 section "Step 5/5: Helm upgrade"
 
--- a/generic/kubernetes/migration/3-cutover.sh
+++ b/generic/kubernetes/migration/3-cutover.sh
@@ -335,6 +335,17 @@ if [[ "${ESTIMATE_MODE}" == "true" ]]; then
     echo ""
     echo "No changes were made to the running application."
     echo "You can run the real cutover with:  ./3-cutover.sh"
+elif [[ "${SKIP_HELM_UPGRADE:-false}" == "true" ]]; then
+    # Data-only migration: the realm/databases have been migrated to the target
+    # backends, but the Helm upgrade is intentionally deferred to the caller
+    # (e.g. a CI harness that owns the chart upgrade to the next version).
+    section "Step 5/5: Helm upgrade — SKIPPED (SKIP_HELM_UPGRADE=true)"
+    save_state "PHASE_3_DURATION" "$(timer_elapsed)"
+    section "Data migration complete ($(timer_elapsed))"
+    echo ""
+    echo "Data has been migrated to the target backends."
+    echo "The Helm upgrade was skipped (SKIP_HELM_UPGRADE=true) — the caller is"
+    echo "responsible for upgrading Camunda to point at the new backends."
 else
 section "Step 5/5: Helm upgrade"
 
--- a/generic/kubernetes/migration/3-cutover.sh
+++ b/generic/kubernetes/migration/3-cutover.sh
@@ -356,8 +356,11 @@ if [[ "${MIGRATE_ELASTICSEARCH}" == "true" ]] && ! is_external_es; then
 fi
 
 if [[ "${MIGRATE_KEYCLOAK}" == "true" ]]; then
-    # Keycloak always uses the operator — values are always needed
-    if [[ -n "${CAMUNDA_DOMAIN:-}" && "${CAMUNDA_DOMAIN}" != "localhost" ]]; then
+    if is_external_keycloak; then
+        # External Keycloak: point Camunda at the pre-existing instance and
+        # disable the bundled Bitnami subchart.
+        HELM_VALUES_ARGS+=("$(generate_external_keycloak_values)")
+    elif [[ -n "${CAMUNDA_DOMAIN:-}" && "${CAMUNDA_DOMAIN}" != "localhost" ]]; then
         HELM_VALUES_ARGS+=("$(get_helm_values keycloak-domain)")
     else
         HELM_VALUES_ARGS+=("$(get_helm_values keycloak-no-domain)")
--- a/generic/kubernetes/migration/3-cutover.sh
+++ b/generic/kubernetes/migration/3-cutover.sh
@@ -161,16 +161,19 @@ do_final_pg_backup() {
     save_state "${component^^}_FINAL_BACKUP" "${component}-db-final.dump"
 }
 
+# The final backup reads the SOURCE database, so it honours the same
+# *_SOURCE_DB_NAME/_USER overrides as the Phase 2 initial backup (defaulting to
+# the target names). The restore below uses the target names.
 if [[ "${MIGRATE_IDENTITY}" == "true" ]]; then
-    do_final_pg_backup identity "${IDENTITY_DB_NAME}" "${IDENTITY_DB_USER}"
+    do_final_pg_backup identity "${IDENTITY_SOURCE_DB_NAME:-${IDENTITY_DB_NAME}}" "${IDENTITY_SOURCE_DB_USER:-${IDENTITY_DB_USER}}"
 fi
 
 if [[ "${MIGRATE_KEYCLOAK}" == "true" ]]; then
-    do_final_pg_backup keycloak "${KEYCLOAK_DB_NAME}" "${KEYCLOAK_DB_USER}"
+    do_final_pg_backup keycloak "${KEYCLOAK_SOURCE_DB_NAME:-${KEYCLOAK_DB_NAME}}" "${KEYCLOAK_SOURCE_DB_USER:-${KEYCLOAK_DB_USER}}"
 fi
 
 if [[ "${MIGRATE_WEBMODELER}" == "true" ]]; then
-    do_final_pg_backup webmodeler "${WEBMODELER_DB_NAME}" "${WEBMODELER_DB_USER}"
+    do_final_pg_backup webmodeler "${WEBMODELER_SOURCE_DB_NAME:-${WEBMODELER_DB_NAME}}" "${WEBMODELER_SOURCE_DB_USER:-${WEBMODELER_DB_USER}}"
 fi
 
 if [[ "${MIGRATE_ELASTICSEARCH}" == "true" ]]; then
--- a/generic/kubernetes/migration/3-cutover.sh
+++ b/generic/kubernetes/migration/3-cutover.sh
@@ -370,11 +370,9 @@ if [[ "${MIGRATE_ELASTICSEARCH}" == "true" ]] && ! is_external_es; then
 fi
 
 if [[ "${MIGRATE_KEYCLOAK}" == "true" ]]; then
-    if is_external_keycloak; then
-        # External Keycloak: point Camunda at the pre-existing instance and
-        # disable the bundled Bitnami subchart.
-        HELM_VALUES_ARGS+=("$(generate_external_keycloak_values)")
-    elif [[ -n "${CAMUNDA_DOMAIN:-}" && "${CAMUNDA_DOMAIN}" != "localhost" ]]; then
+    # External Keycloak is data-only (it implies SKIP_HELM_UPGRADE=true), so this
+    # helm-upgrade branch only ever runs for the operator-managed Keycloak.
+    if [[ -n "${CAMUNDA_DOMAIN:-}" && "${CAMUNDA_DOMAIN}" != "localhost" ]]; then
         HELM_VALUES_ARGS+=("$(get_helm_values keycloak-domain)")
     else
         HELM_VALUES_ARGS+=("$(get_helm_values keycloak-no-domain)")
--- a/generic/kubernetes/migration/README.md
+++ b/generic/kubernetes/migration/README.md
@@ -165,6 +165,7 @@ Set any `MIGRATE_*` to `false` to skip a component (e.g. if it's not deployed or
 | ---------------------------- | --------------- | -------------------------------------------- |
 | `PG_TARGET_MODE`             | `operator`      | `operator`: deploy CNPG. `external`: skip deployment, migrate to pre-existing target |
 | `ES_TARGET_MODE`             | `operator`      | `operator`: deploy ECK. `external`: skip deployment, point to pre-existing target |
+| `KEYCLOAK_TARGET_MODE`       | `operator`      | `operator`: deploy the Keycloak Operator + CR. `external`: skip deployment, point Camunda at a pre-existing Keycloak. Requires `PG_TARGET_MODE=external` |
 
 **Operator mode** (when `*_TARGET_MODE=operator`):
 
@@ -188,7 +189,13 @@ Set any `MIGRATE_*` to `false` to skip a component (e.g. if it's not deployed or
 | `EXTERNAL_ES_PORT`              | `443`                    | Elasticsearch port                           |
 | `EXTERNAL_ES_SECRET`            | `external-es`            | K8s Secret containing the password           |
 | `CUSTOM_HELM_VALUES_FILE`       | (empty)                  | Helm values file for external connections    |
-| `CUSTOM_KEYCLOAK_CONFIG_FILE`   | (empty)                  | Custom Keycloak CR for external PG           |
+| `CUSTOM_KEYCLOAK_CONFIG_FILE`   | (empty)                  | Custom Keycloak CR for external PG (operator mode) |
+| `EXTERNAL_KEYCLOAK_PROTOCOL`    | `http`                   | External Keycloak protocol (`KEYCLOAK_TARGET_MODE=external`) |
+| `EXTERNAL_KEYCLOAK_HOST`        | (empty)                  | External Keycloak host                       |
+| `EXTERNAL_KEYCLOAK_PORT`        | `80`                     | External Keycloak port                       |
+| `EXTERNAL_KEYCLOAK_CONTEXT_PATH`| `/auth`                  | External Keycloak context path               |
+| `EXTERNAL_KEYCLOAK_REALM`       | `/realms/camunda-platform`| External Keycloak realm path                |
+| `EXTERNAL_KEYCLOAK_ADMIN_SECRET`| (empty)                  | K8s Secret with the Keycloak admin `password` |
 
 ### When to use `external` target mode
 
@@ -199,10 +206,16 @@ Set `PG_TARGET_MODE=external` or `ES_TARGET_MODE=external` when the migration **
 | Fresh cluster, no operators installed | `operator` (default) | Scripts install CNPG/ECK + create clusters |
 | Operators already installed by a platform team | `external` | Avoids overwriting the operator version (scripts apply a pinned version via `kubectl apply --server-side`) |
 | Target is a managed service (RDS, OpenSearch, …) | `external` | No operator needed — data migrates directly to the managed endpoint |
+| Keycloak runs as a managed/standalone/Helm instance (not the operator) | `KEYCLOAK_TARGET_MODE=external` | Migrates the realm into the external Keycloak database and points Camunda at the existing instance |
 
 When using `external` mode, you must also provide:
 - Connection details: `EXTERNAL_PG_*` or `EXTERNAL_ES_*` variables in `env.sh`
 - A `CUSTOM_HELM_VALUES_FILE` with Helm values pointing Camunda at the external targets
+
+When using `KEYCLOAK_TARGET_MODE=external`, you must also provide the `EXTERNAL_KEYCLOAK_*`
+variables, and `PG_TARGET_MODE=external` (the external Keycloak serves the migrated realm
+from the external Keycloak database). The scripts do not deploy or manage the Keycloak
+instance itself — only the realm database is migrated and Camunda is rewired to it.
 
 ## Detailed Phase Descriptions
 
--- a/generic/kubernetes/migration/README.md
+++ b/generic/kubernetes/migration/README.md
@@ -217,10 +217,11 @@ When using `external` mode, you must also provide:
 - Connection details: `EXTERNAL_PG_*` or `EXTERNAL_ES_*` variables in `env.sh`
 - A `CUSTOM_HELM_VALUES_FILE` with Helm values pointing Camunda at the external targets
 
-When using `KEYCLOAK_TARGET_MODE=external`, you must also provide:
-- `EXTERNAL_KEYCLOAK_*` connection variables
-- `PG_TARGET_MODE=external` (the external Keycloak serves the migrated realm from the external Keycloak database)
-- `SKIP_HELM_UPGRADE=true` — **external Keycloak mode is data-only**. The scripts migrate the realm database and then exit; the caller must perform the Helm upgrade, wiring Camunda at the external Keycloak (`global.identity.keycloak.url` + auth credentials). The scripts do not wire Keycloak admin credentials into the Helm values.
+When using `KEYCLOAK_TARGET_MODE=external`, you only set that one flag plus the
+`EXTERNAL_KEYCLOAK_*` connection variables. The two coupled settings are derived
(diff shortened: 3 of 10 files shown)


eamonnmoloney:
Added a second commit: `SKIP_HELM_UPGRADE` (data-only cutover). Phase 3 migrates data to the target backends but skips the final `helm upgrade`, so a CI harness can own the N→N+1 chart upgrade onto the migrated external infrastructure. This is what lets the Camunda Helm CI drive these scripts from a pre-upgrade lifecycle hook (migrate data → runner performs the chart upgrade).

leiicamundi:
@eamonnmoloney currently the `main` branch is kind of broken - you can base your branch on `stable/8.9` or wait https://github.com/camunda/camunda-deployment-references/pull/2587 to be merged

leiicamundi:
@eamonnmoloney as discussed, once https://github.com/camunda/camunda-deployment-references/pull/2624 is merged, please rebase this branch onto `stable/8.9` as the migration tooling is going to be removed from `main`.

I'll proceed to the review of the PR after that and you should benefit of the tests pipeline working again

leiicamundi:
### ⚠️ The `migration-tests-heavy` failure can be ignored for this PR

The two failing jobs — `migration-tests-heavy / Kind-Heavy-Domain` and `Kind-Heavy-WarmReindex-Domain` — are **not caused by this PR**.

**Why:** both jobs die at the *"Run benchmark to generate Zeebe data"* step (data seeding), which runs **before** any of the migration logic this PR touches (Phases 1–5 never execute). The job-level annotation is:

> The self-hosted runner lost communication with the server. … starves it for CPU/Memory …

The heavy benchmark (150 PI/s for 30 min, ~20 GB of Elasticsearch data) saturates the self-hosted runner that hosts the kind cluster; the runner runs out of CPU/memory and drops its connection mid-run (logs go silent at ~70% backpressure). The `RESOURCE_EXHAUSTED` / `DEADLINE_EXCEEDED` lines at the start are gateway warm-up transients — the benchmark recovered and was making progress before the runner died. This is infra flakiness, completely independent of `KEYCLOAK_TARGET_MODE` / `SKIP_HELM_UPGRADE`.

The light legs (`Kind-Domain`, `Kind-NoDomain`) pass.

**Root cause + fix:** the heavy legs are flagged `weekly_schedule_only: true` but the matrix filter only excluded `schedule_only` on regular PRs, so they wrongly ran on every PR. I opened #2641 (targeting `stable/8.9`) to exclude `weekly_schedule_only` distros on regular PRs as well — so these heavy legs run only on the weekly Tuesday schedule.

Once #2641 (and #2624, per the rebase discussion above) land, please rebase this branch and the migration pipeline should be green.

leiicamundi:
@eamonnmoloney I'll review tomorrow this PR - I reran https://github.com/camunda/camunda-deployment-references/actions/runs/27210441198?pr=2620 which was failing.

Heavy tests have been removed, so we should get a green test from the pipeline

generic/kubernetes/migration/lib.sh
```diff
@@ -715,6 +725,63 @@ warn_customization() {
...
+        errs=1
+    fi
+
+    if ! is_external_pg; then
+        log_error "KEYCLOAK_TARGET_MODE=external requires PG_TARGET_MODE=external"
+        log_error "  The external Keycloak must serve the migrated realm from the external Keycloak database."
+        errs=1
+    fi
+
+    if [[ -z "${EXTERNAL_KEYCLOAK_ADMIN_SECRET:-}" ]]; then
```
leiicamundi: `EXTERNAL_KEYCLOAK_ADMIN_SECRET` is hard-required here but is referenced **nowhere else** in the
codebase — it never reaches the Helm values. So every caller (incl. camunda-platform-helm#6343) must
create a secret the scripts silently ignore. Either wire it into
`global.identity.keycloak.auth.secret.existingSecret` (see comment on `generate_external_keycloak_values`)
or stop requiring it.

generic/kubernetes/migration/lib.sh
```diff
@@ -715,6 +725,63 @@ warn_customization() {
...
+    fi
+    return "$errs"
+}
+
+# Generate Camunda Helm values that point at an EXTERNAL Keycloak instance.
+# Used during cutover instead of the operator keycloak values. Disables the
+# bundled Bitnami subchart and wires global.identity.keycloak.url at the
+# external endpoint. The admin user override (client tokens + adminUser) is
+# applied separately by sync_keycloak_admin_credentials.
+generate_external_keycloak_values() {
```
leiicamundi: The comment says the admin/client-token override "is applied separately by
`sync_keycloak_admin_credentials`", but that function **early-returns in external mode**
(`lib.sh:1400-1403`), so it's never applied. Combined with `identityKeycloak.enabled: false` (which
stops the chart auto-rendering client-secret env vars), an external-KC migration that lets these
scripts run `helm upgrade` (i.e. `SKIP_HELM_UPGRADE` unset) brings Identity up **without**
connectors/orchestration/console client secrets, and without `global.identity.keycloak.auth`
(adminUser + `EXTERNAL_KEYCLOAK_ADMIN_SECRET`).

Fix options: (a) require `SKIP_HELM_UPGRADE=true` when `KEYCLOAK_TARGET_MODE=external` and document it
as data-only; or (b) emit `global.identity.keycloak.auth.{adminUser,secret→EXTERNAL_KEYCLOAK_ADMIN_SECRET}`
here and run the client-token/override logic in external mode (skipping only the operator-secret
reconcile, which the inner guard at `lib.sh:1432` already handles).

generic/kubernetes/migration/lib.sh
```diff
@@ -1315,6 +1392,16 @@ deploy_keycloak() {
 # we must update the operator secret to match the restored DB credentials.
 # Also generates a Helm override file to set the correct adminUser.
 sync_keycloak_admin_credentials() {
+    # Operator-only reconcile: it updates the operator's "keycloak-initial-admin"
+    # secret and renders an admin override for the Helm upgrade. In external mode
+    # there is no operator secret, the external Keycloak's admin comes from the
+    # restored realm database, and the caller wires Camunda's admin credentials
+    # (and, with SKIP_HELM_UPGRADE, owns the upgrade and its values). So skip it.
+    if is_external_keycloak; then
```
leiicamundi, line 1383: This blanket early-return is the root of the incomplete external path: it skips **both** the
operator-only secret reconcile **and** the still-needed client-token patching + admin override
generation (lines 1446-1521). Note the inner `if ! is_external_keycloak … else …` at lines 1432-1442
is now **dead code** (unreachable once we've returned here). Prefer dropping this early-return and
letting the inner guard skip only the operator secret, so the override is still produced for external
mode.

generic/kubernetes/migration/2-backup.sh
```diff
@@ -72,16 +72,22 @@ do_pg_backup() {
...
+# source names from the target names (which Phase 3 restore uses); they default
+# to the target names for the common same-name case.
 if [[ "${MIGRATE_IDENTITY}" == "true" ]]; then
-    do_pg_backup identity "${IDENTITY_DB_NAME}" "${IDENTITY_DB_USER}"
+    do_pg_backup identity "${IDENTITY_SOURCE_DB_NAME:-${IDENTITY_DB_NAME}}" "${IDENTITY_SOURCE_DB_USER:-${IDENTITY_DB_USER}}"
 fi
 
 if [[ "${MIGRATE_KEYCLOAK}" == "true" ]]; then
-    do_pg_backup keycloak "${KEYCLOAK_DB_NAME}" "${KEYCLOAK_DB_USER}"
+    do_pg_backup keycloak "${KEYCLOAK_SOURCE_DB_NAME:-${KEYCLOAK_DB_NAME}}" "${KEYCLOAK_SOURCE_DB_USER:-${KEYCLOAK_DB_USER}}"
```
leiicamundi, line 86: These `*_SOURCE_DB_NAME` / `*_SOURCE_DB_USER` overrides are required for real Bitnami sources
(camunda-platform-helm#6343 sets `KEYCLOAK_SOURCE_DB_NAME=bitnami_keycloak`,
`KEYCLOAK_SOURCE_DB_USER=bn_keycloak`), but they are not declared in `env.sh` nor documented in
`README.md`. Every other knob in this tool is discoverable via `env.sh` — please add
`IDENTITY/KEYCLOAK/WEBMODELER_SOURCE_DB_NAME/_USER` (defaulting to the target names) with a short
comment + a README row.

generic/kubernetes/migration/env.sh
```diff
@@ -82,6 +82,23 @@ export ES_INDEX_PREFIXES="${ES_INDEX_PREFIXES:-zeebe-* operate-* tasklist-* opti
...
 
+# Keycloak target mode:
+#   "operator"  → Scripts deploy the Keycloak Operator + a Keycloak CR.
+#   "external"  → Scripts do NOT deploy Keycloak. An external Keycloak instance
+#                 (managed service, companion Helm release, standalone) must
+#                 already exist and be backed by the external Keycloak database
+#                 (EXTERNAL_PG_KEYCLOAK_*), so the restored realm is served by it.
+#                 Configure the "External Keycloak instance" subsection below.
+# Requires PG_TARGET_MODE=external (the external Keycloak owns its database).
+export KEYCLOAK_TARGET_MODE="${KEYCLOAK_TARGET_MODE:-operator}"
```
leiicamundi, line 94: Two small follow-ups around this new block:
1. Declare the `*_SOURCE_DB_NAME/_USER` overrides here too (they're used in 2-backup.sh / 3-cutover.sh
   but undocumented).
2. The "Migration config loaded" summary echo at the bottom of this file prints PG/ES target modes —
   please add `KEYCLOAK_TARGET_MODE` and `SKIP_HELM_UPGRADE` so the active mode is visible in CI logs.

generic/kubernetes/migration/jobs/pg-backup.job.yml
```diff
@@ -21,6 +21,14 @@ spec:
                 migration.camunda.io/type: backup
         spec:
             restartPolicy: Never
+            # Make the backup PVC writable by the non-root postgres image
+            # (UID/GID 1001 in the Bitnami/community images). Without an fsGroup
+            # the dynamically-provisioned volume is root-owned and `mkdir
+            # /backup/<component>` fails with EACCES on storage classes that
+            # don't pre-chown the mount (e.g. GKE pd.csi).
+            securityContext:
+                fsGroup: 1001
```
leiicamundi, line 30: Nit: `fsGroup: 1001` assumes the Bitnami/community PG image UID — correct for this migration-from-
Bitnami scenario, just coupled to `PG_IMAGE`. Good fix for GKE `pd.csi`. 👍

generic/kubernetes/migration/README.md
```diff
@@ -199,11 +206,25 @@ Set `PG_TARGET_MODE=external` or `ES_TARGET_MODE=external` when the migration **
...
+from the external Keycloak database). The scripts do not deploy or manage the Keycloak
+instance itself — only the realm database is migrated and Camunda is rewired to it.
+
+### Advanced: data-only cutover (`SKIP_HELM_UPGRADE`)
+
+Set `SKIP_HELM_UPGRADE=true` to run Phase 3 data migration (backup/restore +
+reindex) but skip the final `helm upgrade`. The caller then owns the chart
+upgrade. This is intended for CI harnesses that migrate Bitnami data onto
+external infrastructure and then perform an N→N+1 chart upgrade themselves
+(for example, the Camunda Helm test suites). Normal migrations leave it `false`.
```
leiicamundi, line 239: Please also document, in this external-mode section: (1) that `KEYCLOAK_TARGET_MODE=external`
currently targets the data-only / `SKIP_HELM_UPGRADE=true` flow (caller owns the upgrade), and (2) the
`*_SOURCE_DB_NAME/_USER` overrides needed when the source Bitnami DB/user names differ from the target
(e.g. `bitnami_keycloak`/`bn_keycloak`).

leiicamundi (changes requested):
## Review summary

Solid, well-structured change. The operator path is fully preserved and the **CI use case this
actually enables** (`KEYCLOAK_TARGET_MODE=external` + `SKIP_HELM_UPGRADE=true`, caller owns the chart
upgrade — see camunda-platform-helm#6343) is correct. Local lint is fully green and the light CI legs
pass.

**Verified locally:** `bash -n`, `shellcheck --exclude=SC2155 --external-sources`, and
`yamllint` are all clean on the changed files; no trailing-whitespace / final-newline / exec-bit
issues. So the "shellcheck/pre-commit not run" caveat in the description can be considered cleared.

### Main point (🔴) — the external-Keycloak-with-`helm upgrade` path is incomplete
When `KEYCLOAK_TARGET_MODE=external` **without** `SKIP_HELM_UPGRADE`, the scripts run the helm upgrade
themselves but never wire Keycloak credentials: `generate_external_keycloak_values()` omits
`global.identity.keycloak.auth`, `sync_keycloak_admin_credentials()` early-returns (so the
client-token override is never generated), and `identityKeycloak.enabled: false` stops the chart from
auto-rendering client secrets. `EXTERNAL_KEYCLOAK_ADMIN_SECRET` is even required+validated but never
consumed. CI is green because #6343 uses `SKIP_HELM_UPGRADE=true` (helm-upgrade branch skipped). See
inline comments. Recommend either gating external KC to `SKIP_HELM_UPGRADE=true` (+document) or
completing the wiring.

### Should-fix (🟡)
- `*_SOURCE_DB_NAME/_USER` are required for real Bitnami sources (#6343 sets
  `KEYCLOAK_SOURCE_DB_NAME=bitnami_keycloak`) but undocumented in `env.sh`/`README.md`.
- `warn_customization()` still prints "Keycloak Operator IS still deployed" in external mode.
- `env.sh` startup summary doesn't echo `KEYCLOAK_TARGET_MODE` / `SKIP_HELM_UPGRADE`.

### Docs
No companion docs PR exists yet; the 8.9 guide currently states managed Keycloak is *not* assumed,
which this PR changes. A docs PR is being prepared separately.

_Full report attached in the team channel; happy to pair on the wiring._

leiicamundi:
📝 Companion docs PR (draft, **8.9 versioned only**): camunda/camunda-docs#9054

It documents `KEYCLOAK_TARGET_MODE=external` + `EXTERNAL_KEYCLOAK_*`, the `SKIP_HELM_UPGRADE` data-only flow, and the `*_SOURCE_DB_NAME/_USER` overrides, and updates the "Keycloak is not managed" statement in the Bitnami → managed-services guide. It's intentionally held as draft until this PR lands, since the documented contract (the `SKIP_HELM_UPGRADE` requirement and whether `EXTERNAL_KEYCLOAK_ADMIN_SECRET` is actually wired) depends on how the review comments here are resolved.

generic/kubernetes/migration/lib.sh
eamonnmoloney: Fixed in 4f5a1d6. You're right — it was validated but never consumed (no Helm values path reads it), so every caller was creating a secret the scripts ignored.

Rather than wire it in, I dropped `EXTERNAL_KEYCLOAK_ADMIN_SECRET` entirely. External KC is data-only: the realm (incl. admin) comes out of the `pg_restore`, and with `SKIP_HELM_UPGRADE=true` the caller owns the upgrade and supplies `global.identity.keycloak.auth` itself (#6343 does exactly this). The validation now enforces that data-only contract instead of requiring an unused secret.

generic/kubernetes/migration/lib.sh
eamonnmoloney: Addressed in 4f5a1d6 by choosing your option (a): gate external KC to the data-only flow rather than completing the `helm upgrade` wiring.

`validate_external_keycloak_config` now hard-requires `SKIP_HELM_UPGRADE=true`, and as of 1e66dde sourcing `env.sh` derives that automatically from `KEYCLOAK_TARGET_MODE=external` (along with `PG_TARGET_MODE=external`), so the customer sets one flag. Because the scripts never run the upgrade in this mode, the missing `global.identity.keycloak.auth` / client-secret wiring is the caller's responsibility (and is owned by #6343). The stale "applied separately by sync_keycloak_admin_credentials" comment no longer describes a path the scripts take.

generic/kubernetes/migration/lib.sh
eamonnmoloney: Good catch on the dead code. Since we went with option (a) (external KC is data-only, `SKIP_HELM_UPGRADE=true`), the blanket early-return is actually the correct behaviour now — the override file is irrelevant because the scripts never run the upgrade in this mode, and the caller produces those values.

The dead inner guard you flagged at line 1432 is removed in 78494bd: the reconcile now runs unconditionally (we're always in operator mode past the early return), and the unreachable `else` branch is gone.

generic/kubernetes/migration/2-backup.sh
eamonnmoloney: Done in 4f5a1d6. All six `IDENTITY/KEYCLOAK/WEBMODELER_SOURCE_DB_NAME/_USER` overrides are now declared in `env.sh` (defaulting to the target `*_DB_NAME`/`*_DB_USER`), with a comment explaining the Bitnami name mismatch, and a README row documenting the `bitnami_keycloak`/`bn_keycloak` defaults.

generic/kubernetes/migration/env.sh
eamonnmoloney: Both done in 4f5a1d6:
1. The `*_SOURCE_DB_NAME/_USER` overrides are now declared in `env.sh` (defaulting to the target names).
2. The "Migration config loaded" summary now echoes `KC target:` and `Skip upgrade:` alongside the existing PG/ES lines, so the active mode is visible in CI logs.

generic/kubernetes/migration/jobs/pg-backup.job.yml
eamonnmoloney: Thanks — agreed, leaving `fsGroup: 1001` as-is since it's coupled to `PG_IMAGE` (the Bitnami source) by design for this migration scenario.

generic/kubernetes/migration/README.md
eamonnmoloney: Done in 4f5a1d6, and simplified further in 1e66dde. The external-mode section now documents:
1. That `KEYCLOAK_TARGET_MODE=external` is the data-only flow — and that `SKIP_HELM_UPGRADE=true` + `PG_TARGET_MODE=external` are now **derived automatically** when you source `env.sh` (a `[auto]` line is logged for each), so the customer sets one flag rather than three.
2. The `*_SOURCE_DB_NAME/_USER` overrides for when the source Bitnami DB/user names differ (`bitnami_keycloak`/`bn_keycloak`), with an example block.

eamonnmoloney:
Thanks for the thorough review @leiicamundi — all points addressed. Summary:

**RED — external-KC-with-`helm upgrade` was incomplete.** Resolved via your option (a): external KC is now an explicit data-only mode.
- Dropped `EXTERNAL_KEYCLOAK_ADMIN_SECRET` (validated-but-unused — it never reached the Helm values).
- `validate_external_keycloak_config` enforces the data-only contract; the caller owns the upgrade and the `global.identity.keycloak.auth` wiring (as #6343 does).
- Removed the dead inner guard in `sync_keycloak_admin_credentials` (unreachable past the early return).

**YELLOWs:**
- `*_SOURCE_DB_NAME/_USER` now declared in `env.sh` + documented in `README.md` (incl. the `bitnami_keycloak`/`bn_keycloak` defaults).
- `warn_customization()` no longer prints "Keycloak Operator IS still deployed" in external mode; an external-KC info block is shown instead.
- Startup summary now echoes `KC target:` and `Skip upgrade:`.

**Bonus simplification (1e66dde):** following the "set it internally" thread, `KEYCLOAK_TARGET_MODE=external` now **auto-derives** its two coupled settings — `PG_TARGET_MODE=external` and `SKIP_HELM_UPGRADE=true` — when you source `env.sh`, logging a `[auto]` line for each. The customer sets one flag plus the `EXTERNAL_KEYCLOAK_*` connection vars instead of keeping three in lock-step. The lib.sh guards remain as a defensive safety net for post-source overrides.

`bash -n` clean on all touched scripts. Ready for another look.

leiicamundi (approved):
## Re-review ✅ — all points addressed

Thanks for the fast turnaround. All seven items from my previous review are resolved, and the design
is cleaner than what I proposed: external Keycloak is now explicitly **data-only**, with the coupled
flags **auto-derived** from a single `KEYCLOAK_TARGET_MODE=external` (loud `[auto]` log), so a user
sets one flag instead of three.

**Resolution:**
- 🔴 *External KC + auto `helm upgrade` incomplete* → gated to data-only: `validate_external_keycloak_config` now requires `SKIP_HELM_UPGRADE=true`, and `env.sh` auto-derives `PG_TARGET_MODE=external` + `SKIP_HELM_UPGRADE=true` (`4f5a1d62`, `1e66dde2`).
- 🔴 *`EXTERNAL_KEYCLOAK_ADMIN_SECRET` required but unused* → removed entirely (env.sh, README, validation); no residual refs.
- 🟡 *`*_SOURCE_DB_NAME/_USER` undocumented* → declared in `env.sh` + documented in the README table with the Bitnami `bitnami_keycloak`/`bn_keycloak` defaults.
- 🟡 *`warn_customization` misleading* → gated + dedicated external-mode banner.
- 🟡 *`env.sh` summary* → now prints `KC target:` and `Skip upgrade:`.
- 🟢 *dead `else` in `sync_keycloak_admin_credentials`* → removed (`78494bd3`).
- 🟢 *README `SKIP_HELM_UPGRADE` constraint* → documented.

**Follow-up nit — handled directly:** since external KC now always implies `SKIP_HELM_UPGRADE=true`,
`generate_external_keycloak_values()` and the external branch in `3-cutover.sh` became unreachable (and
would have re-introduced the missing-auth-wiring bug only via a contrived `env.sh` bypass). I pushed a
small cleanup commit (`8b4f81e3`, `refactor(migration): drop unreachable external-Keycloak helm-values
path`) removing them — shout if you'd rather revert it. The helm-upgrade path now only handles the
operator Keycloak.

**Verified on `8b4f81e3`:** `bash -n` + `shellcheck --exclude=SC2155 --external-sources` clean; the
pre-existing light CI legs (`Kind-Domain`, `Kind-NoDomain`, `lint / pre-commit`) were green on the
previous head and CI will re-run on the cleanup commit before merge.

Companion docs PR camunda/camunda-docs#9054 (draft, 8.9) already matches this final contract; I'll flip
it out of draft once this merges. LGTM. ✅

generic/kubernetes/migration/lib.sh
```diff
@@ -1315,6 +1392,16 @@ deploy_keycloak() {
...
+        log_info "External Keycloak — skipping operator admin-credential sync"
+        return 0
+    fi
+
     log_info "Syncing Keycloak admin credentials from Bitnami → operator ..."
```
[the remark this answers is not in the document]
leiicamundi: Resolved. External Keycloak is now strictly **data-only**: `KEYCLOAK_TARGET_MODE=external` auto-implies `SKIP_HELM_UPGRADE=true` (derived in `env.sh`, enforced by `validate_external_keycloak_config`), so the scripts never run the in-cluster Helm upgrade in external mode — the caller owns it and wires the client secrets. The function that set `identityKeycloak.enabled=false` (`generate_external_keycloak_values`) was also removed in 8b4f81e3, so the early return here is now correct.

generic/kubernetes/migration/lib.sh
```diff
@@ -715,6 +725,63 @@ warn_customization() {
...
+        keycloak:
+            internal: false
+            url:
+                protocol: ${EXTERNAL_KEYCLOAK_PROTOCOL:-http}
+                host: ${EXTERNAL_KEYCLOAK_HOST}
+                port: "${EXTERNAL_KEYCLOAK_PORT:-80}"
+            contextPath: ${EXTERNAL_KEYCLOAK_CONTEXT_PATH:-/auth}
+            realm: ${EXTERNAL_KEYCLOAK_REALM:-/realms/camunda-platform}
+identityKeycloak:
+    enabled: false
```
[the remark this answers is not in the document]
leiicamundi: Resolved. `EXTERNAL_KEYCLOAK_ADMIN_SECRET` was removed entirely (no longer validated or required), and `generate_external_keycloak_values` was removed in 8b4f81e3. External Keycloak is data-only now, so Camunda↔Keycloak auth (admin secret + client tokens) is wired by the caller during its deferred Helm upgrade, not by these scripts.

generic/kubernetes/migration/3-cutover.sh
```diff
@@ -335,6 +338,17 @@ if [[ "${ESTIMATE_MODE}" == "true" ]]; then
...
+    # Data-only migration: the realm/databases have been migrated to the target
+    # backends, but the Helm upgrade is intentionally deferred to the caller
+    # (e.g. a CI harness that owns the chart upgrade to the next version).
+    section "Step 5/5: Helm upgrade — SKIPPED (SKIP_HELM_UPGRADE=true)"
+    save_state "PHASE_3_DURATION" "$(timer_elapsed)"
+    section "Data migration complete ($(timer_elapsed))"
+    echo ""
+    echo "Data has been migrated to the target backends."
+    echo "The Helm upgrade was skipped (SKIP_HELM_UPGRADE=true) — the caller is"
+    echo "responsible for upgrading Camunda to point at the new backends."
```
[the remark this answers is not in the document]
leiicamundi: Good catch — fixed in 8c0ea9af. The `SKIP_HELM_UPGRADE` branch now calls `run_hooks post-phase-3` + `complete_phase 3` so the phase state machine is consistent and `4-validate.sh`/`5-cleanup-bitnami.sh` are no longer blocked. The output now also makes explicit that Camunda stays frozen on the old backends until the caller's Helm upgrade, so those later phases should run only after that switch.

generic/kubernetes/migration/lib.sh
```diff
@@ -217,11 +217,21 @@ check_env() {
...
+                    # previous minor — the caller upgrades to ${expected_version}
+                    # after the data has moved onto the external backends. (This is
+                    # the only supported order when the target minor has dropped the
+                    # bundled Bitnami subcharts, so the source cannot be on it.)
+                    log_warn "Deployed Camunda version ${deployed_app_version} (${deployed_major_minor}) differs from ${expected_version}; allowed because SKIP_HELM_UPGRADE=true (caller performs the version upgrade)"
+                else
+                    log_error "Deployed Camunda version ${deployed_app_version} (${deployed_major_minor}) does not match expected version ${expected_version} from .camunda-version"
+                    log_error "These migration scripts are designed for Camunda ${expected_version}. Check that you are using the correct branch."
+                    exit 1
+                fi
```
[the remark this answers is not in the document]
leiicamundi: Acknowledged — leaving as-is for now. The data-only callers (CI harness) pin the source branch, and enforcing 'exactly the previous minor' can be too rigid across patch/minor combinations. Noting it as a possible future hardening rather than changing behavior in this PR.

leiicamundi:
After the rebase onto the latest `stable/8.9` (head is now `4fe84246`), the migration test workflow was still running on the previous head (`8c0ea9af`) and — because the workflow's concurrency group uses `cancel-in-progress: false` — it was blocking the run for the current commit.

I cancelled that stale run so the workflow for `4fe84246` could start. We're now waiting on the migration integration tests for the current head to go green: https://github.com/camunda/camunda-deployment-references/actions/runs/27338203617

The rebase left the `generic/kubernetes/migration/` tree byte-identical to the previously reviewed/approved state, so the approval still stands.

leiicamundi:
Heads-up: the merge was blocked by the `stable/8.9` branch protection (**Require signed commits**) — all 14 commits in the branch were unsigned (the rebase force-push had stripped signatures). The repo only allows squash-merge.

To unblock, I rebased the branch and **re-signed all 14 commits** with my GPG key (registered on my account, so they now show **Verified**), then force-pushed. The tree is byte-identical to the previous head `4fe84246` — only the signatures and the committer field changed; **authorship is preserved** (your commits still show you as author).

I also cancelled the stale workflow run on the previous head so the run on the re-signed head (`6cc548f8`) can proceed. Approval still stands; once CI is green this should be mergeable. @eamonnmoloney let me know if you'd prefer to re-sign under your own key instead.

PR #2623, opened 2026-06-05
feat(eks): move ingress controller to Contour

marcel-dias:
## Description

Forward-port of #2276 (opened against `stable/8.9`) to `main`. Migrates the **EKS single-region** reference architecture and its **IRSA** variant from `ingress-nginx` to **Project Contour**.

### Motivation

`main` diverged from `stable/8.9` since #2276 was branched, so this PR applies the migration's *net* changes adapted to `main` (notably, `main` already moved `host` to a top-level `global.host`, so the `className: contour` keys are anchored on `ingress.enabled` rather than the former `ingress.host`).

### What changed

- **Install scripts:** replace `install-ingress-nginx.sh` with `install-contour.sh` for `eks-single-region` (and update the `eks-single-region-irsa` symlink). Azure and other ref-archs are untouched and keep nginx.
- **Helm values (`values-domain.yml`, both EKS variants):** set `global.ingress.className: contour` and `orchestration.ingress.grpc.className: contour`, and add `projectcontour.io/upstream-protocol.h2c: '26500'` on the orchestration `Service` so Zeebe gRPC (h2c) works through Contour.
- **Chart version:** add `CONTOUR_HELM_CHART_VERSION` (with a Renovate annotation) to `export-ingress-setup-vars.sh`.
- **`aws-kubernetes-ingress-setup` action:** add a selectable `ingress-controller` input (**`nginx` stays the default**, `contour` opt-in) with a script-existence guard.
- **`kubernetes-ingress-cleanup` action:** add Contour uninstall + `projectcontour` namespace cleanup.
- **`kubernetes-keycloak-operator` action:** add a `domain-contour` mode mapping to the existing `keycloak-instance-domain-contour.yml`.
- **CI:** wire the EKS single-region and operator-based integration tests to use Contour.

### Intentional differences from #2276

To keep `main` aligned with active development, this backport **does not** carry two changes from the stable PR:

- The Helm deprecation check stays **strict** (no `continue-on-error`; the `global.opensearch.enabled` exclude-pattern is retained).
- The `generic/kubernetes/single-region/tests/integration/**` workflow trigger path is **kept** — those tests still exist and are active on `main`.

refs #2103


leiicamundi:

---

## Review — `feat(eks): move ingress controller to Contour` (#2623, backport of #2276 → `main`)

This is the forward-port of #2276 to `main`, and the adaptation is well done — anchoring `className: contour` on `ingress.enabled` (since `main` already moved `host` to top-level `global.host`) is exactly right, and I appreciate that the intentional differences from the stable PR are spelled out in the description. I verified the same correctness points as on #2276; they hold here too.

### What I verified ✅

- **Value paths valid** against `charts/camunda-platform-8.9/values.yaml`: `global.ingress.className`, `orchestration.ingress.grpc.className`, `orchestration.service.annotations` all documented. The bot's "unknown key" flag on `...service.annotations.projectcontour.io/upstream-protocol.h2c` is a **false positive** (free-form annotations map).
- **Contour `0.4.0`** creates a cluster-default `IngressClass` named `contour` → `className: contour` binds correctly.
- **Keeping the deprecation check strict** here (no `continue-on-error`, retaining the `global.opensearch.enabled` exclude-pattern) is the right call for `main`, and the inline comment explaining the legacy Operate OpensearchConnector dependency is helpful.
- **Keeping the `generic/kubernetes/single-region/tests/integration/**` trigger path** is correct — those tests are still active on `main`.
- Diff is otherwise identical to #2276 (same action wiring, cleanup, keycloak `domain-contour` mode, `install-contour.sh`).

### Suggestions (same as #2276)

1. **`ingressClass.default: true` — add a documentation warning.** Contour's chart marks its `IngressClass` cluster-default. Fine on a fresh, dedicated cluster, but on an **existing** cluster it can hijack any `Ingress` without an explicit `className` (e.g. ones previously handled by nginx) and break the running setup. Recommend a short **`> [!WARNING]`** in the docs so operators don't apply it to an existing cluster unawares — documentation only, no chart change needed.
2. **`nginx` remains the action input default while `install-ingress-nginx.sh` is deleted for this ref-arch.** Guarded correctly; a sentence in the action README on per-ref-arch availability would avoid confusion.

### Cross-repo coordination ⚠️

The docs PR camunda/camunda-docs#8989 updates **both** current and 8.9 versioned docs, pulling `install-contour.sh` live via `reference` blocks:
- current docs → `blob/main/.../install-contour.sh` (needs **this PR #2623** merged to `main`)
- 8.9 docs → `blob/stable/8.9/.../install-contour.sh` (needs **#2276** merged to `stable/8.9`)

So #8989 must merge only **after both** #2623 and #2276 land, or the docs build will 404 on the missing script. Recommend coordinating the three so this PR + #2276 go in before the docs PR.

### Consistency check

This backport faithfully mirrors #2276 with the documented, justified deltas. No divergence concerns. Once the doc-comment nits above are addressed (and ideally kept in sync with #2276), this LGTM. 👍

aws/kubernetes/eks-single-region/helm-values/values-domain.yml
```diff
@@ -104,13 +105,18 @@ orchestration:
...
             host: zeebe-${CAMUNDA_DOMAIN}
             tls:
                 enabled: true
                 secretName: zeebe-c8-tls-grpc
             annotations:
                 kubernetes.io/tls-acme: 'true'
 
+    service:
+        annotations:
+            projectcontour.io/upstream-protocol.h2c: '26500'
```
leiicamundi, line 118: ✅ Verified `orchestration.service.annotations` is a valid, documented key in `camunda-platform-8.9` (free-form map), so the deprecation bot's “unknown key” flag on this annotation is a false positive. h2c on 26500 is the correct mechanism for Zeebe gRPC through Contour/Envoy.

aws/kubernetes/eks-single-region/procedure/install-contour.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+helm upgrade --install \
+  contour contour \
+  --repo https://projectcontour.github.io/helm-charts/ \
```
leiicamundi, line 6: Contour chart `0.4.0` sets `ingressClass.{create:true,default:true}`, so it creates a **cluster-default** `IngressClass contour`. On an **existing** cluster this can hijack any `Ingress` without an explicit `className` and break the running setup. Recommend a `> [!WARNING]` in the docs about applying this to an existing cluster (docs-only — no chart change needed).

leiicamundi (commented):

aws/kubernetes/eks-single-region/procedure/install-contour.sh
marcel-dias: set default to false

PR #2641, opened 2026-06-08
fix(ci): exclude weekly_schedule_only distros from regular PR runs

leiicamundi:
## Motivation

The `Tests - Integration - Generic Migration from Bitnami` workflow fails on **every regular pull request** at the data-seeding step, before any migration logic runs. This currently blocks functional PRs such as #2620.

### Root cause

The two heavy benchmark legs in the migration matrix are flagged `weekly_schedule_only: true`:

- `Kind-Heavy-Domain`
- `Kind-Heavy-WarmReindex-Domain`

But `internal-tests-matrix/action.yml` only excluded `schedule_only` on regular PRs — **not** `weekly_schedule_only`. So the heavy legs ran on every PR, even though they are meant to run only on the weekly Tuesday cron.

The heavy benchmark (`benchmark-job-heavy.yml`: 150 PI/s for 30 min, ~20 GB of Elasticsearch data) saturates the self-hosted runner that hosts the kind cluster. The runner runs out of CPU/memory and drops its connection mid-run:

> The self-hosted runner lost communication with the server. … Anything in your workflow that terminates the runner process, **starves it for CPU/Memory**, or blocks its network access can cause this error.

The job is killed during step *"Run benchmark to generate Zeebe data"* (~70% backpressure, runner dies), so the migration phases never even execute.

## What this PR does

Aligns the regular-PR filter with the intent of `weekly_schedule_only`: distros flagged `weekly_schedule_only: true` are now excluded on regular PRs (same as they already are for renovate PRs). They still run on the weekly schedule.

| Trigger | Heavy legs run? |
| --- | --- |
| Scheduled (Tuesday cron) | ✅ yes (unchanged) |
| Regular PR | ❌ no (**fixed** — was incorrectly yes) |
| Renovate PR | ❌ no (unchanged) |

Only the migration matrix uses `weekly_schedule_only`, so no other workflow is affected.

## Validation

- `yamllint -c .lint/yamllint/.yamllint.yaml` passes (lines stay under the 175 limit).
- Verified the `yq` filter against the real matrix file:
  - Regular PR → `Kind-NoDomain`, `Kind-Domain` only.
  - Renovate PR → `Kind-NoDomain`, `Kind-Domain` (unchanged).
  - Scheduled → all four legs including both heavy (unchanged).
- pre-commit hooks (yamllint, yamlfmt, action README) pass.

## Related

- Unblocks #2620 (the heavy-leg failure there is this infra issue, not a defect in that PR).

### DIFF
--- a/.github/actions/internal-tests-matrix/action.yml
+++ b/.github/actions/internal-tests-matrix/action.yml
@@ -66,7 +66,11 @@ runs:
                 platform_matrix="$(yq '(.matrix |= (.distro |= map(select(.weekly_schedule_only == null or .weekly_schedule_only == false)))) | .matrix' \
                   --indent=0 --output-format json "${{ inputs.ci_matrix_file }}")"
               else
-                platform_matrix="$(yq '(.matrix |= (.distro |= map(select(.schedule_only == null or .schedule_only == false)))) | .matrix' \
+                echo "This is a regular PR, we exclude schedule_only and weekly_schedule_only scenarios."
+                platform_matrix="$(yq '(.matrix |= (.distro |= map(select(
+                    (.schedule_only == null or .schedule_only == false)
+                    and (.weekly_schedule_only == null or .weekly_schedule_only == false)
+                  )))) | .matrix' \
                   --indent=0 --output-format json "${{ inputs.ci_matrix_file }}")"
               fi
 


leiicamundi:
## CI failures on this PR are unrelated infra issues — not caused by the change

This PR only edits the shared `internal-tests-matrix` action, so it fans out to **every** cloud test workflow (that action is in their path filters). The 3 red checks are pre-existing infra flakiness, none related to the matrix-filter change. Details below, plus a root-cause analysis of the VPN quota one as requested.

### 1. ROSA single-region `Prepare clusters` — AWS Client VPN quota exhausted

The ROSA cluster itself created fine (`Apply complete! Resources: 67 added`). The failure is the **VPN** stage:

```
Error: creating EC2 Client VPN Endpoint: ... api error ClientVpnEndpointLimitExceeded: Limit exceeded
  with module.vpn.aws_ec2_client_vpn_endpoint.vpn,
  on ../../../../modules/vpn/endpoint.tf line 25
```

**Why the quota is full even though cleanup exists** — I traced the whole lifecycle:

- Two ref-archs provision a Client VPN endpoint in **eu-west-2**: `aws/openshift/rosa-hcp-single-region` and `aws/kubernetes/eks-single-region`. AWS's **default soft quota is 5 Client VPN endpoints per Region**.
- The single-region test workflows use `concurrency: cancel-in-progress: false`. A shared-action PR like this one fans out to **both** VPN-creating ref-archs across **multiple matrix legs** (this run alone: rosa-single ×1 + eks-single `domain` + `irsa` = 3 endpoints), and those runs all stay alive concurrently — on top of other open PRs, the `*_golden` workflows, and the daily scheduled runs. The 5-slot regional quota is exceeded, and the *next* endpoint to be created is the one that gets rejected.
- The **per-run** cleanup is correct (`aws-generic-terraform-cleanup`, `modules-order: vpn,cluster`, `if: always()`), but it only frees the slot **at the end** of each run, and Client VPN teardown is slow (network-association disassociation + `aws_ec2_client_vpn_authorization_rule` has a 20-min delete timeout). So each endpoint keeps occupying a quota slot for many minutes after the test is logically done, overlapping with newly starting runs.
- The **daily** cleanup (`aws_rosa_hcp_single_region_daily_cleanup.yml`) is **age-gated** (`max_age_hours: 12`) and **tfstate-based**. Its `delete-ghost-rosa-clusters` flag only sweeps orphaned **ROSA/OCM clusters** — there is **no equivalent sweeper for orphaned Client VPN endpoints**. So any endpoint left behind by a partial-failure run (VPN created, tfstate not fully persisted / runner killed) lingers until someone deletes it by hand.

So it's a **concurrency-vs-soft-quota collision**, amplified by slow VPN teardown and the absence of a ghost-VPN sweeper — not a missing cleanup.

Remediation options (separate from this PR):
1. Request a **Service Quotas** increase for "Client VPN endpoints per Region" in eu-west-2 (default 5 → higher).
2. Add a **ghost Client VPN endpoint sweeper** to the daily cleanup (delete CI-tagged endpoints older than N hours regardless of tfstate), mirroring `delete-ghost-rosa-clusters`.
3. Reduce simultaneous VPN usage (share one endpoint across matrix legs, or don't fan out every VPN-creating cloud workflow on shared-action PRs).

### 2 & 3. EKS single-region `🔥 Run Camunda Smoke Test` (both `domain` and `irsa`)

The `kubectl port-forward` to the Zeebe gateway drops ~30s into the create loop and is never re-established (only `146/1485` instances created, then all 30 verify attempts return `HTTP 000000` — a client-side connection failure, not a Camunda error; the empty-filter diagnostic probe is also `000`). The smoke script even documents that it can't restart the port-forward itself.

I'm opening a separate fix PR to make the port-forward **self-healing** (supervisor auto-restart on SPDY-stream drop) — link to follow.

---

The migration-heavy fix itself is working as intended here: `migration-tests-heavy` is now **skipped** and `Kind-Domain` / `Kind-NoDomain` pass. `required_status_checks` on `stable/8.9` is empty, so these unrelated red checks don't block the merge.

PR #2658, opened 2026-06-11
fix(migration): exclude transient JGROUPS_PING from the Keycloak realm dump

eamonnmoloney:
## Problem

The Bitnami→external Keycloak realm migration (`generic/kubernetes/migration`) does a full `pg_dump`/`pg_restore` of the Keycloak database, which **includes `JGROUPS_PING`** — Keycloak's JDBC_PING *cluster-discovery* table. Its rows hold the **source** cluster's node addresses. Restored into a fresh target Keycloak, they collide with the target's startup membership insert:

```
ERROR: Failed to start server in (production) mode
JDBC exception executing SQL [INSERT INTO JGROUPS_PING values (?, ?, ?, ?, ?)]
duplicate key value violates unique constraint "constraint_jgroups_ping"
  Detail: Key (address)=(uuid://...0002) already exists.
```

→ the migrated Keycloak `CrashLoopBackOff`s and never serves the restored realm.

## Fix

`backup_pg()` now passes `--exclude-table-data=jgroups_ping` for the **keycloak** component (via a new optional `PG_DUMP_EXTRA_ARGS` that the backup job appends to `pg_dump`). The table **schema** is preserved (restored empty); only the transient cluster-membership rows are dropped, so the target re-registers its own membership on boot. Identity/WebModeler dumps are unaffected — the variable is empty for them.

Restore already does `DROP SCHEMA public CASCADE` + `pg_restore --clean` under `FORCE_RESTORE=true`, so the target gets the dump's empty `jgroups_ping` and starts clean.

## How it was found

Discovered while validating the camunda-platform-helm **post-infra migration hook** (camunda/camunda-platform-helm#6343) on a real **8.9→8.10 modular-upgrade** on GKE. Every data-plane step succeeded — external-target detection, Bitnami Keycloak PG backup, ES warm reindex (9 indices), Phase 3 cutover, **realm restore confirmed** (`camunda-platform` + `master` realms present in the companion Keycloak DB). The carried-over `JGROUPS_PING` row was the **sole** blocker.

## Validation

`bash -n` clean; `envsubst` smoke test confirms the keycloak `pg_dump` gets the flag and other components stay unaffected. Re-running the full GKE 8.9→8.10 migration against this branch to confirm the migrated Keycloak now starts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/generic/kubernetes/migration/README.md
+++ b/generic/kubernetes/migration/README.md
@@ -477,3 +477,29 @@ aws route53 change-resource-record-sets \
         "ResourceRecords":[{"Value":"keycloak-service.camunda.svc.cluster.local"}]
     }}]}'
 ```
+
+## Troubleshooting
+
+### Migrated Keycloak crashes on startup (`constraint_jgroups_ping`)
+
+If the migrated Keycloak enters `CrashLoopBackOff` with a startup error such as:
+
+```text
+ERROR: Failed to start server in (production) mode
+JDBC exception executing SQL [INSERT INTO JGROUPS_PING values (?, ?, ?, ?, ?)]
+duplicate key value violates unique constraint "constraint_jgroups_ping"
+  Detail: Key (address)=(uuid://...0002) already exists.
+```
+
+the realm dump carried over rows from `JGROUPS_PING`, Keycloak's transient JDBC_PING
+cluster-discovery table. Those rows hold the source cluster's node addresses and
+collide with the target's own startup membership insert.
+
+The migration scripts handle this automatically: `backup_pg` excludes the
+`JGROUPS_PING` table **data** (keeping its schema) from the Keycloak realm dump, so
+the target starts clean and re-registers its own membership. This applies to both
+`KEYCLOAK_TARGET_MODE=operator` and `KEYCLOAK_TARGET_MODE=external`.
+
+If you hit this on a custom or older pipeline, exclude the table data from the dump
+(`pg_dump --exclude-table-data=jgroups_ping …`) or `TRUNCATE jgroups_ping` in the
+target database before starting Keycloak.
--- a/generic/kubernetes/migration/jobs/pg-backup.job.yml
+++ b/generic/kubernetes/migration/jobs/pg-backup.job.yml
@@ -2,6 +2,8 @@
 # PostgreSQL Backup Job (generic, parameterized by COMPONENT)
 # Required env vars: COMPONENT, JOB_NAME, PG_HOST, PG_PORT, PG_DATABASE,
 #   PG_USERNAME, PG_SECRET_NAME, PG_SECRET_KEY, PG_IMAGE, BACKUP_PVC, NAMESPACE, TIMESTAMP
+# Optional: PG_DUMP_EXTRA_ARGS — extra pg_dump flags (e.g. --exclude-table-data
+#   to drop transient tables that must not cross clusters; see lib.sh backup_pg).
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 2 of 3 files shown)


leiicamundi:
TODO: update doc (fyi ai reviewer)

eamonnmoloney:
ℹ️ **Note on the force-push** (`2e8b74e` → `5af5adc`): content-only no-op — the commit was **amended solely to add an SSH signature** so it satisfies this repo's signed-commits merge requirement (the original was pushed before commit signing was configured locally). The tree is byte-identical; `git diff 2e8b74e 5af5adc` is empty. No code, message, or author changes. The commit now shows **Verified**.

leiicamundi (commented):
## Code review — JGROUPS_PING exclusion

Reviewed the fix end to end against the migration scripts (`lib.sh`, `jobs/pg-backup.job.yml`, `jobs/pg-restore.job.yml`, `env.sh`). **The approach is correct.** Excluding the `JGROUPS_PING` *data* (keeping its schema) is the right way to stop the transient cluster-discovery rows from crossing clusters, and the restore path (`pg_restore --clean --if-exists` + the `FORCE_RESTORE=true` `DROP SCHEMA public CASCADE`) restores the table empty so the target re-registers its own membership.

Validation I ran on the branch:
- `bash -n lib.sh` clean; `shellcheck 0.11.0 -S warning` clean; `pre-commit` passes on the changed files.
- `envsubst` render check: the `keycloak` component receives the exclude flag(s) and other components do not.
- Confirmed the unquoted `${PG_DUMP_EXTRA_ARGS}` word-splits correctly under `/bin/bash` (the job's interpreter) into separate arguments.

### Follow-up I pushed (0af731fb)

1. **Case robustness** — also pass `--exclude-table-data=JGROUPS_PING` (uppercase). PostgreSQL folds unquoted identifiers to lowercase, so `jgroups_ping` covers the normal case (and matches the observed `constraint_jgroups_ping`), but if any build created the table with a quoted identifier the lowercase pattern would silently no-op. A non-matching `--exclude-table-data` pattern is never an error (independent of `--strict-names`), so this is safe insurance.
2. **Comment clarity** — noted the exclusion protects both `operator` and `external` Keycloak target modes (both restore the realm dump into a fresh Keycloak; the crash surfaces on the next Keycloak start).
3. **README troubleshooting** — added a "Migrated Keycloak crashes on startup (`constraint_jgroups_ping`)" entry documenting the symptom and the automatic handling.

### Remaining gap (not addressed here — needs a validated PR)

The migration test matrix (`.github/workflows-config/.../test_matrix.yml`) is **operator-mode only** — `_migration_test_run.yml` hardcodes `PG_TARGET_MODE=operator`, and there is no `KEYCLOAK_TARGET_MODE=external` entry. This is exactly the path where the bug surfaced, so CI cannot catch a regression today. Adding an external-Keycloak (companion-infra) matrix entry is the proper regression guard, but it requires real CI validation and companion wiring, so I did not bolt it on blind. Recommend tracking it as a follow-up (it was already noted as a follow-up in #2620).

_Docs companion updated in camunda/camunda-docs#9054._

leiicamundi:
Follow-up on the external-Keycloak CI coverage I flagged above: for the `external` target mode (`KEYCLOAK_TARGET_MODE=external` / `PG_TARGET_MODE=external`) we rely on the **distro/QA test suites**, which drive these migration scripts in `external` mode end to end. This repo's migration matrix intentionally stays operator-only on kind, so the external path is exercised by distro/QA rather than here — no dedicated external entry is needed in this matrix.

leiicamundi (approved):
@eamonnmoloney lgtm, let's wait for the latest commits to trigger the tests run then you can merge

PR #2665, opened 2026-06-12
fix(ci): skip identityKeycloak IRSA check when using operator Keycloak

leiicamundi:
## Description

The AWS IRSA check (`internal-camunda-chart-tests` → `c8-sm-checks/checks/kube/aws-irsa.sh`) hardcoded `identityKeycloak` in the PostgreSQL component list:

```bash
PG_COMPONENTS="identityKeycloak,identity"
```

In the `eks-single-region-irsa` scenario, Keycloak is deployed via the **Keycloak Operator** (`keycloak-service:18080`, with its own `pg-keycloak-*` database), so the bundled `identityKeycloak` (Bitnami) subchart is disabled and never deployed. The check then fails every `identityKeycloak.*` lookup and `aws-irsa.sh` exits `163`:

```
[FAIL] identityKeycloak.externalDatabase.host is not defined in your helm values.
[FAIL] The Aurora host is not set. ...
[FAIL] The service account camunda-identityKeycloak does not have a valid eks.amazonaws.com/role-arn annotation.
[FAIL] ./.c8-sm-checks/checks/kube/aws-irsa.sh: At least one of the checks failed (error code: 163).
```

This has been a **standing failure** of the scheduled `Tests - Integration - AWS Kubernetes EKS Single Region (IRSA)` workflow on `main` (failing weekly since the IRSA scenario moved to the Keycloak Operator in #1630). All `[FAIL]` lines are `identityKeycloak`-only; `identity` (PostgreSQL) and `orchestration`/`optimize` (OpenSearch) IRSA checks already pass.

## Change

Only include `identityKeycloak` in the IRSA PostgreSQL component list when **no** external/operator Keycloak is configured. The action already receives `keycloak-service-name` (set to `keycloak-service:18080` only in operator mode, empty otherwise), so it is used as the signal.

- Operator/external Keycloak → `PG_COMPONENTS="identity"` (+`webModeler` if enabled)
- Bundled Keycloak → `PG_COMPONENTS="identityKeycloak,identity"` (**unchanged behaviour**)

No new inputs, no caller changes.

## Backport

- `stable/8.9` — affected (same operator wiring), backport PR to follow.
- `stable/8.8` / `8.7` / `8.6` — **not affected**: 8.8's IRSA scenario uses the bundled Keycloak (so the check is correct there) and 8.7/8.6 don't have this code path.

## Verification

- `bash -n` + `shellcheck` on the embedded logic; runtime check confirms both branches produce the expected component lists.
- `pre-commit` content hooks (yamllint, yamlfmt, check-yaml, …) pass. Docker-based hooks (actionlint/README) unavailable locally — covered by the `lint / pre-commit` CI job.

### DIFF
--- a/.github/actions/internal-camunda-chart-tests/action.yml
+++ b/.github/actions/internal-camunda-chart-tests/action.yml
@@ -473,6 +473,7 @@ runs:
           env:
               NAMESPACE: ${{ inputs.test-namespace }}
               ELASTICSEARCH_ENABLED: ${{ inputs.elasticsearch-enabled }}
+              KEYCLOAK_SERVICE_NAME: ${{ inputs.keycloak-service-name }}
           run: |
               set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


leiicamundi:
Consolidating to reduce open PRs: the identityKeycloak IRSA fix is already included in #2651 (same base `main`), which carries this exact change (commit 29608dc0) alongside the JWKS-warmup fix. Closing in favour of #2651 so the fix lands there. Reopen if we decide to keep it standalone.

PR #2689, opened 2026-06-17
feat(aks-single-region): migrate ingress controller to Contour

marcel-dias:
## Description

Migrates the **AKS single region** reference architecture (and its `-rdbms` variant) from **ingress-nginx** to **[Project Contour](https://github.com/projectcontour/contour)**, mirroring the EKS migration in #2276.

[ingress-nginx was retired in March 2026](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/), so all reference architectures must move to a maintained ingress controller. The shared Contour infrastructure (chart version, keycloak `domain-contour` mode, ingress cleanup) already landed on `main` in #2623; this PR wires AKS to use it.

## Changes

- **`procedure/install-contour.sh`** (new) replaces `install-ingress-nginx.sh`. Installs Contour from the official upstream chart into the `projectcontour` namespace, translating the previous Azure-specific settings onto the Envoy service:
  - `envoy.service.externalTrafficPolicy=Local` (preserves client source IP and lets the Azure LB health probe use the Kubernetes `healthCheckNodePort`)
  - `service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path=/healthz`
  - `contour.replicaCount=2` to keep the HA parity the nginx install had
  - The `-rdbms` variant symlinks to the same script (as it did for nginx).
- **Helm values** (`values-domain.yml`, both variants):
  - `className: contour` on `global.ingress` and `orchestration.ingress.grpc`
  - `projectcontour.io/upstream-protocol.h2c: '26500'` service annotation so Zeebe gRPC works through Envoy
- **`azure-kubernetes-ingress-setup` action**: installs Contour instead of ingress-nginx, cleans the `projectcontour` namespace, updated description + regenerated README.
- **AKS test workflow**: keycloak operator switched to `domain-contour`; the Helm deprecation check is temporarily `continue-on-error` (matching #2276) until the chart schema aligns with the new Contour values.

## Implications

- `no-domain` scenarios are unaffected (no ingress).
- No Terraform/golden-plan changes — this is a Helm/procedure/CI change only.
- End-to-end coverage comes from the `Tests - Integration - Azure AKS Single Region` workflow (domain declination): Contour install → cert-manager TLS → external-dns Azure DNS records → Camunda reachable via domain → Zeebe gRPC reachable.

refs: https://github.com/camunda/team-infrastructure-experience/issues/1074
refs: https://github.com/camunda/team-infrastructure-experience/issues/933


.github/workflows/azure_kubernetes_aks_single_region_tests.yml
```diff
@@ -685,7 +685,9 @@ jobs:
 
                   ./generic/kubernetes/single-region/procedure/install-chart.sh
 
+            # TODO: re-enable once Helm chart schema aligns with the new Contour values
             - name: 🔍 Check for Helm deprecation warnings
+              continue-on-error: true
```
hisImminence, line 692: **P2 · `continue-on-error: true` on already-non-blocking deprecation check swallows real tool failures** _(via devils-advocate)_

The `internal-helm-deprecation-check` action is designed to be non-blocking for deprecation warnings — it posts PR comments and Slack alerts but does not exit non-zero for warnings. The action only exits non-zero on genuine internal tool failures (e.g., `helm get notes` auth error, schema validator crash). Adding `continue-on-error: true` on top of this already-non-blocking action means real execution errors will be silently swallowed, removing the signal that the check ran at all. The TODO comment references 'Helm chart schema aligns with the new Contour values' as the exit condition but provides no linked issue or automated tracking, making this effectively permanent by default. Recommendation: if the action supports an `exclude-patterns` input, use it to suppress Contour-specific schema warnings while preserving tool-failure signals. If not, open a tracking issue and link it directly in the TODO comment so the exception has an explicit owner and won't be forgotten.

```
              continue-on-error: true
```

hisImminence (commented):
**crev review**



_Specialists run: correctness, devils-advocate. Devil's-advocate hypotheses: 0 raised, 0 promoted._

---

**Findings on lines outside this PR's diff:**

- **P2** `azure/kubernetes/aks-single-region/helm-values/values-domain.yml:22` — Nginx chart-default annotations deep-merged onto Contour Ingress objects
  The Camunda Helm chart sets nginx-specific ingress annotation defaults (e.g. `nginx.ingress.kubernetes.io/ssl-redirect: "false"`, `nginx.ingress.kubernetes.io/proxy-buffer-size: "128k"`, `nginx.ingress.kubernetes.io/proxy-buffering: "on"`, `nginx.ingress.kubernetes.io/proxy-body-size: "10m"`). Because the values file only adds `kubernetes.io/tls-acme: 'true'` without explicitly clearing those defaults, Helm's deep-merge will include all nginx-specific annotations on the rendered Contour-managed Ingress objects. The gRPC ingress (line 115 of the same file) is additionally subject to `nginx.ingress.kubernetes.io/backend-protocol: GRPC` appearing in the rendered manifest. Contour silently ignores unknown nginx annotations so functionality is unaffected, but the rendered manifests carry dead annotations that will confuse operators. The same issue applies to `azure/kubernetes/aks-single-region-rdbms/helm-values/values-domain.yml` line 25. To fix, explicitly null out the nginx defaults in both files:
  
  ```yaml
  global:
    ingress:
      annotations:
        nginx.ingress.kubernetes.io/ssl-redirect: null
        nginx.ingress.kubernetes.io/proxy-buffer-size: null
        nginx.ingress.kubernetes.io/proxy-buffering: null
        nginx.ingress.kubernetes.io/proxy-body-size: null
  ```
  
  Same pattern for `orchestration.ingress.grpc.annotations` (line 115), clearing `nginx.ingress.kubernetes.io/backend-protocol: null` in addition.

.github/actions/azure-kubernetes-ingress-setup/action.yml
```diff
@@ -63,10 +63,10 @@ runs:
           with:
               namespace: external-dns
 
-        - name: 🏗️ Clean ingress-nginx operator namespace
+        - name: 🏗️ Clean Contour operator namespace
           uses: ./.github/actions/internal-clean-namespace
           with:
-              namespace: ingress-nginx
+              namespace: projectcontour
```
hisImminence, line 69: why not just contour?

hisImminence (approved):
Thank you! Only one non blocking question and two P2s to iterate over.

leiicamundi (changes requested):
As described in https://github.com/camunda/camunda-deployment-references/issues/2691, let's remove completely ingress-nginx from the repo as aks is the last one requiring it

azure/kubernetes/aks-single-region/procedure/install-contour.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+helm upgrade --install \
+  contour contour \
+  --repo https://projectcontour.github.io/helm-charts/ \
+  --version "$CONTOUR_HELM_CHART_VERSION" \
+  --set 'contour.replicaCount=2' \
```
leiicamundi, line 12: **[B3 — blocker] Align `contour.ingressClass.default` with EKS.** Please add `--set 'contour.ingressClass.default=false'` alongside these flags to match `aws/kubernetes/eks-single-region/procedure/install-contour.sh`. Without it the chart default (`true`) marks Contour as the **cluster-default IngressClass** on AKS, which diverges from EKS. Every Camunda Ingress sets `className: contour` explicitly so it works either way, but let's keep the two providers aligned.

azure/kubernetes/aks-single-region/procedure/install-contour.sh
```diff
@@ -0,0 +1,12 @@
...
+  --set 'envoy.service.externalTrafficPolicy=Local' \
```
leiicamundi, line 13: **[minor] `Local` is already the chart default** for `envoy.service.externalTrafficPolicy` (EKS omits this flag). Fine to keep it explicit since the Azure LB health-probe annotation on the next line depends on `Local` (so the probe uses the Kubernetes `healthCheckNodePort`) — but please add a one-line comment saying so, otherwise it reads as redundant vs EKS.

azure/kubernetes/aks-single-region/helm-values/values-domain.yml
```diff
@@ -15,6 +15,7 @@ global:
     host: ${CAMUNDA_DOMAIN}
     ingress:
         enabled: true
+        className: contour
```
leiicamundi, line 18: **[B2 — chart-side] cc @hisImminence.** The Camunda chart deep-merges nginx-specific annotation **defaults** onto these Contour-managed Ingress objects: `global.ingress.annotations` (`nginx.ingress.kubernetes.io/ssl-redirect`, `proxy-buffer-size`, `proxy-buffering`, `proxy-body-size`) and `orchestration.ingress.grpc.annotations` (`nginx.ingress.kubernetes.io/backend-protocol: GRPC`, …). Since these files only **add** `kubernetes.io/tls-acme`, Helm keeps the nginx defaults and the rendered Ingress carries dead nginx annotations (Contour ignores them, but the manifests are misleading).

Now that ingress-nginx is retired, these look like dead defaults that should be dropped from **camunda-platform-helm upstream**, rather than null-ed out in every Contour ref-arch. @hisImminence — you flagged this on the Ingress side; can you confirm/route the chart-side removal? I'd rather not bake nginx-specific null-outs into the Contour values if the chart fix is coming, so not asking @marcel-dias to patch this here.

.github/workflows/azure_kubernetes_aks_single_region_tests.yml
leiicamundi, line 692: **[minor] Keeping `continue-on-error` as-is** (team decision). For the record: `internal-helm-deprecation-check` is non-blocking by design — findings go to a PR comment, not a job failure — and CI confirms the new `projectcontour.io/upstream-protocol.h2c` key surfaces only as a non-blocking unknown-keys warning, so this flag does not change pass/fail today. Optional later: link #2691 in the adjacent TODO so the exception is tracked, and prefer an `exclude-patterns:` entry over `continue-on-error` so genuine tool failures (e.g. `helm get notes` errors) aren't swallowed.

leiicamundi (changes requested):
Thanks @marcel-dias — the AKS→Contour migration itself is correct and faithfully mirrors the EKS reference: values parity (`className: contour` on global + gRPC ingress, `projectcontour.io/upstream-protocol.h2c: '26500'`), the gRPC h2c wiring, the `domain-contour` keycloak mode, `no-domain` untouched, and the `-rdbms` symlink all check out. `CONTOUR_HELM_CHART_VERSION` is sourced before the install and the Contour value keys are valid.

Requesting changes on one repo-side point, with a second routed to the chart team:

**[B1 — blocker] Remove ingress-nginx from the repo entirely (#2691).** AKS was the last reference architecture using it, so let's not leave the repo half-migrated. Suggested scope (this PR or a coordinated follow-up tracked by #2691):
- `generic/kubernetes/single-region/procedure/export-ingress-setup-vars.sh` — drop `INGRESS_HELM_CHART_VERSION` + its renovate line.
- `.github/actions/aws-kubernetes-ingress-setup/` — the `ingress-controller` input and its `nginx` branch are already dead code (`aws/kubernetes/eks-single-region/procedure/install-ingress-nginx.sh` no longer exists, so selecting `nginx` hits the "install script not found" guard); make the action Contour-only + regenerate the README.
- `.github/actions/kubernetes-ingress-cleanup/` — drop the `ingress-nginx` uninstall + namespace-deletion steps and the description/README mention.
- `local/kubernetes/kind-single-region/` — `ingress-nginx-deploy.sh` and the `coredns-configmap.yaml` rewrite to `ingress-nginx-controller.ingress-nginx.svc...` (a `contour-deploy.sh` + `coredns-configmap-contour.yaml` already exist to switch to).
- `aws/kubernetes/eks-dual-region/procedure/manifests/nginx.yml` and the keycloak `domain` (nginx) mode / `keycloak-instance-domain-nginx.yml` if no longer referenced, plus the stale `ingress-nginx` comment in `aws-generic-terraform-cleanup`.

**[B3 — blocker]** `contour.ingressClass.default` alignment with EKS — see inline comment.
**[B2 — chart-side]** dead nginx annotation defaults leaking onto the Contour Ingress — see inline comment, cc @hisImminence (the fix belongs upstream in camunda-platform-helm).

**Follow-up (separate PR):** align EKS up to `contour.replicaCount=2` for Contour control-plane HA parity (data-plane HA is already covered by the Envoy DaemonSet on both sides).

Minor inline nits on the install script and on the deprecation-check `continue-on-error` (keeping the latter as-is, as discussed).

marcel-dias:
Thanks for the thorough review! Addressed below.

### ✅ Fixed in this PR
- **B3 — `contour.ingressClass.default=false`** (align with EKS): added in `install-contour.sh`.
- **minor — `externalTrafficPolicy=Local`**: kept explicit with a one-line comment explaining the Azure LB health-probe annotation depends on it (probe targets the Kubernetes `healthCheckNodePort`).
- **deprecation-check `continue-on-error`**: kept as agreed; the adjacent TODO now links the upstream tracking issue **camunda/camunda-platform-helm#4564** (the unknown-key — `projectcontour.io/upstream-protocol.h2c` — is a chart-schema gap, which is exactly the re-enable condition).
- **golden plans (CI red)**: the `azurerm_key_vault_key` `release_policy` diff was azurerm v4.77 provider drift (pre-existing on `main`); regenerated/aligned both golden files. Both `compare` jobs are green now.

### ➡️ Moved to a dedicated follow-up — #2717
- **B1 — remove ingress-nginx repo-wide**: done in **#2717** (resolves #2691). It strips the controller from the AWS ingress-setup/cleanup actions, the keycloak `domain` (nginx) mode, kind, and `INGRESS_HELM_CHART_VERSION`. It's marked **to merge after this PR**, since AKS is the last consumer of that variable.
  - Note: `aws/kubernetes/eks-dual-region/.../nginx.yml` is intentionally **kept** — it's a `sample-nginx` HTTP pod for the DNS-chaining test, not the ingress controller.

### 💬 Notes
- **B2 — dead nginx annotation defaults on Contour Ingress** (cc @hisImminence): agreed this belongs upstream in camunda-platform-helm rather than null-ing out in each ref-arch, so not patched here.
- **@hisImminence "why not just contour?"**: it is — the Azure action is Contour-only now. The namespace stays `projectcontour` to match the EKS install and the shared `kubernetes-ingress-cleanup` action.

azure/kubernetes/aks-single-region/procedure/install-contour.sh
marcel-dias: Done in ccdaaef3 — added `--set 'contour.ingressClass.default=false'` to `install-contour.sh`, so AKS now mirrors the EKS reference and no longer marks Contour as the cluster-default IngressClass. Thanks for catching this!

azure/kubernetes/aks-single-region/procedure/install-contour.sh
marcel-dias: Added in ccdaaef3 — there's now a comment above the `helm upgrade` call explaining that `envoy.service.externalTrafficPolicy=Local` is kept explicit because the Azure LB health-probe annotation on the next line depends on it (with `Local` the probe targets the Kubernetes `healthCheckNodePort`).

.github/workflows/azure_kubernetes_aks_single_region_tests.yml
marcel-dias: Agreed that `exclude-patterns` would be the cleaner fix so genuine tool failures aren't swallowed. For now we're keeping `continue-on-error: true` as a team decision — `internal-helm-deprecation-check` is non-blocking by design (findings go to a PR comment, not a job failure). The adjacent TODO now links the upstream tracking issue camunda/camunda-platform-helm#4564, which is the explicit exit condition (the chart recognising the new Contour value keys). Switching to `exclude-patterns` is a good follow-up but out of scope for this PR.

marcel-dias: Kept `continue-on-error` as discussed. The adjacent TODO now links camunda/camunda-platform-helm#4564 rather than #2691 — that issue is the actual exit condition (the chart recognising the Contour value keys), whereas #2691 tracks the ingress-nginx removal. Agree `exclude-patterns` is the better long-term approach; leaving that as a follow-up.

.github/actions/azure-kubernetes-ingress-setup/action.yml
marcel-dias: `projectcontour` is the namespace Contour installs into here (`--namespace projectcontour` in `install-contour.sh`), and it matches both the Contour project's conventional namespace and the EKS reference. We can rename it to `contour`, but only as a coordinated change that also aligns EKS and kind — so I'd rather keep it consistent across providers for now and handle the rename separately to avoid divergence.

azure/kubernetes/aks-single-region/helm-values/values-domain.yml
marcel-dias: Leaving these as-is in this PR per your note — the fix belongs upstream in camunda-platform-helm rather than null-ing the nginx defaults in every Contour ref-arch. @hisImminence, over to you to confirm/route the chart-side removal of the dead nginx annotation defaults.

marcel-dias:
**Re: [B1] remove ingress-nginx repo-wide** — handled in the dedicated follow-up #2717 (`refactor(ingress): remove ingress-nginx repo-wide after Contour migration`), tracked by #2691. It covers the full list flagged here: `generic/.../export-ingress-setup-vars.sh`, `.github/actions/aws-kubernetes-ingress-setup`, `.github/actions/kubernetes-ingress-cleanup`, the kind `ingress-nginx-deploy.sh` + coredns rewrite, and the eks-dual-region nginx manifest + keycloak nginx mode. Keeping this PR scoped to the AKS→Contour migration, as you suggested (this PR or a coordinated follow-up).

azure/kubernetes/aks-single-region/helm-values/values-domain.yml
leiicamundi: @hisImminence — I logged the chart-side fix as camunda/camunda-platform-helm#6410, with a reproduction showing the nginx annotation defaults rendering onto an `ingressClassName: contour` Ingress. Could you pick it up / route it so these dead `nginx.ingress.kubernetes.io/*` defaults are dropped from the 8.10 chart? That clears this for every Contour consumer at once instead of null-ing them out per reference architecture. Thanks!

leiicamundi (approved):
Approving — thanks for the quick turnaround @marcel-dias.

- **B3** (`contour.ingressClass.default=false`, EKS parity) — done in `ccdaaef3`.
- **`externalTrafficPolicy=Local`** — now documented inline (Azure LB health-probe dependency).
- **deprecation-check `continue-on-error`** — kept by agreement; the TODO now links the real exit condition (camunda/camunda-platform-helm#4564).
- **B1** (repo-wide ingress-nginx removal) — split into #2717; good call keeping this PR scoped to the AKS→Contour migration. Please merge #2717 right after this, since it removes the now-last `INGRESS_HELM_CHART_VERSION` consumer. And agreed that `aws/kubernetes/eks-dual-region/.../nginx.yml` is a `sample-nginx` test pod (not the controller) — correct to keep it.
- **B2** (dead nginx annotation defaults rendered onto the Contour Ingress) — not a deployment-references fix; logged upstream as camunda/camunda-platform-helm#6410 and handed to @hisImminence.

One non-blocking follow-up parked on #2717: EKS `contour.replicaCount=2` for Contour control-plane HA parity with AKS.

LGTM.

leiicamundi:
⚠️ **Docs sequencing — this PR will break the published AKS docs build on merge.**

The AKS install guide pulls its install step live from this repo via a `reference` block:

- `docs/.../azure/microsoft-aks/aks-helm.md` (current, on `main`) → `https://github.com/camunda/camunda-deployment-references/blob/main/azure/kubernetes/aks-single-region/procedure/install-ingress-nginx.sh`

This PR **deletes** `install-ingress-nginx.sh` from `main`, so as soon as it merges that `reference` block 404s and the camunda-docs build fails (the same failure mode handled for EKS in camunda/camunda-docs#8989).

**Needed before/with this merge:** an AKS docs PR mirroring camunda/camunda-docs#8989 — swap `microsoft-aks/aks-helm.md` from `install-ingress-nginx.sh` → `install-contour.sh` plus the nginx→Contour prose — merged in lockstep with this PR.

Note: the 8.9 AKS docs are safe for now (the Contour migration is not backported to `stable/8.9`, so `install-ingress-nginx.sh` still exists there); only the current/`main` AKS doc is affected. If/when this is backported to `stable/8.9`, the 8.9 AKS doc will need the same update.

cc @marcel-dias

PR #2704, opened 2026-06-18
feat(aks): replace ingress-nginx with Contour on AKS single-region

hisImminence:
## Summary

- Replaces ingress-nginx (retired March 2026) with [Contour](https://projectcontour.io/) as the default ingress controller for `azure/kubernetes/aks-single-region`
- Adds `install-contour.sh`: deploys Contour with `LoadBalancer`, `replicaCount=2`, `externalTrafficPolicy=Local`, and the Azure Standard LB health-probe annotation (`service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path=/healthz`)
- Adds `values-contour-overlay.yml`: sets `ingressClassName: contour` for both HTTP and gRPC ingresses; adds `projectcontour.io/upstream-protocol.h2c: "26500"` to the Zeebe service so Envoy treats the gRPC upstream as HTTP/2 cleartext
- Updates `docs/manual-deployment-azure-aks.md` and `AGENTS.md` to reflect Contour as default

## Zeebe gRPC approach

Contour uses a Service annotation to determine the upstream protocol. Setting `projectcontour.io/upstream-protocol.h2c: "26500"` on the Zeebe gateway service tells Envoy to use h2c (HTTP/2 cleartext) to the pod while still terminating TLS externally. This is equivalent to what nginx achieves via `nginx.ingress.kubernetes.io/backend-protocol: GRPC`.

## Tested on

AKS single-region cluster (`swedencentral`, Kubernetes 1.35, Contour 0.6.0 / Envoy 1.33.5). All web applications (`/`, `/managementidentity/`, `/optimize/`) return expected OIDC redirects via Envoy. gRPC endpoint returns 415 on plain HTTP with valid TLS (expected — Envoy rejects non-gRPC traffic on the h2c upstream).

## Test plan

- [ ] `helm template` with `values-domain.yml` + `values-contour-overlay.yml` — confirm both Ingresses have `ingressClassName: contour`
- [ ] Confirm `camunda-zeebe-gateway` service has `projectcontour.io/upstream-protocol.h2c: "26500"` annotation
- [ ] Deploy to AKS with `install-contour.sh` and verify Envoy LB IP assigned
- [ ] Smoke test: `curl -sI https://<domain>` returns `HTTP/2 302`


hisImminence:
Splitting into separate PRs: Contour implementation will be handled in a follow-up aligned with #2689; agent context docs moved to #2705.

PR #2707, opened 2026-06-18
feat(ci): tag Kubernetes-provisioned NLBs with map-migrated for AWS MAP

leiicamundi:
## What

Tag the **Kubernetes-provisioned** Network Load Balancers in the EKS reference architectures with the AWS MAP tag `map-migrated=migARUADZHVWZ` — applied **in CI only**, so the doc-rendered procedures stay free of Camunda's MAP id.

- `.github/actions/aws-kubernetes-ingress-setup/action.yml` — after installing the ingress controller, annotate its Service (`contour-envoy`/`projectcontour` or `ingress-nginx-controller`/`ingress-nginx`) with `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags`.
- `aws/kubernetes/eks-dual-region/test/internal/helpers/aws/helpers.go` — the dual-region test harness annotates the internal CoreDNS `internal-dns-lb` Service.

The in-tree AWS cloud provider then propagates the tag onto the NLB it creates.

## Why

Camunda is enrolled in the AWS Migration Acceleration Program (MAP), which credits eligible AWS spend on resources tagged `map-migrated = migARUADZHVWZ`. Elastic Load Balancing is eligible.

Everywhere else the tag is delivered through the Terraform AWS provider `default_tags` (injected by CI), which covers every Terraform-managed resource (KMS keys, Terraform-managed ALBs/NLBs, …). The two NLBs above are **not** created by Terraform — they are provisioned by the in-tree AWS cloud provider from a Kubernetes `Service`, so they never inherit `default_tags` and were the only untagged AWS load balancers.

## Why in CI and not in the procedures

The `install-contour.sh` / `install-ingress-nginx.sh` scripts and the `internal-dns-lb.yml` manifest are rendered verbatim in the customer-facing documentation. Hardcoding `map-migrated=migARUADZHVWZ` there would bake Camunda's MAP id into copies users deploy in their own accounts. Instead, the tag is applied from CI:
- the `aws-kubernetes-ingress-setup` action is CI-only (not doc-rendered);
- the dual-region tagging lives in the Go test harness (`test/`), also not doc-rendered.

## Scope / non-goals

- **KMS:** no change needed — all `aws_kms_key` resources are Terraform-managed and already inherit `map-migrated` via `default_tags`.
- **AWS Load Balancer Controller:** not used in the repo; the in-tree annotation is the correct mechanism.
- **ROSA/OpenShift:** uses Routes / the platform router LB managed out of band.

## Backports

- `stable/8.9` — same Contour + dual-region shape (#2708).
- `stable/8.8` (#2709), `stable/8.7` (#2710) — still use `ingress-nginx` and have no `eks-dual-region`, so only the ingress-setup action is patched there.

relates to camunda/team-infrastructure-experience#1151

### DIFF
--- a/.github/actions/aws-kubernetes-ingress-setup/action.yml
+++ b/.github/actions/aws-kubernetes-ingress-setup/action.yml
@@ -144,6 +144,24 @@ runs:
               echo "Installing ingress controller '${{ inputs.ingress-controller }}' via $install_script ..."
               "$install_script"
 
+              # Tag the ingress controller load balancer for the AWS MAP program.
+              # The NLB is provisioned by the in-tree AWS cloud provider from the Service
+              # object, so it does not inherit the Terraform provider default_tags that
+              # carry the map-migrated tag. The tag is applied here, in CI only, so the
+              # reference procedures (which are rendered in the docs) stay free of
+              # Camunda's MAP id; the cloud provider then propagates it onto the NLB.
+              map_lb_service=""
+              map_lb_namespace=""
+              case "${{ inputs.ingress-controller }}" in
+                  nginx)   map_lb_service="ingress-nginx-controller"; map_lb_namespace="ingress-nginx" ;;
+                  contour) map_lb_service="contour-envoy";            map_lb_namespace="projectcontour" ;;
+              esac
+              if [[ -n "$map_lb_service" ]]; then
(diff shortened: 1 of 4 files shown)


leiicamundi:
Backports opened:
- `stable/8.9` → #2708 (Contour + internal-dns-lb, identical patch)
- `stable/8.8` → #2709 (ingress-nginx variant)
- `stable/8.7` → #2710 (ingress-nginx variant)

PR #2736, opened 2026-06-19
fix(ci): wire keycloak wait into EKS/AKS/ROSA-single domain deploys

leiicamundi:
**Stacked on #2735** (which is stacked on #2718). Merge after both.

## What
Makes the in-`install-chart.sh` Keycloak wait (#2735) effective in CI for the single-region domain tests:
- **EKS / AKS**: add `KEYCLOAK_WAIT_INSECURE` to the deploy step (keyed off `cert-config.use-wildcard-cert`), bump the deploy `timeout-minutes` to fit the bounded wait.
- **ROSA-single**: same, with `KEYCLOAK_WAIT_INSECURE: 'true'` (ROSA always uses the Vault wildcard / internal CA).
- **EKS**: remove the now-redundant inline `⏳ Wait for Keycloak external URL reachable` step (its poll-then-rollout-restart logic now lives in `wait-for-keycloak.sh`, called by `install-chart.sh`).

## Why insecure in CI
The CI public cert is the Vault wildcard (internal CA); the readiness probe must skip TLS verification (matches EKS's old `curl -k`).

## Not included
- **ROSA dual-region**: uses **basic auth** (`values-base.yml`: *Management Identity (Keycloak) is not deployed in dual-region*) — no public OIDC issuer, so the crash-loop can't occur there; no keycloak-wait needed. (Its flake is Submariner — separate.)

Validated: `actionlint` clean, `yamllint`/`yamlfmt` pass; EKS/AKS/ROSA single-region CI deploy via `install-chart.sh` + set `CAMUNDA_DOMAIN`, so the wait runs.

### DIFF
--- a/.github/workflows/aws_kubernetes_eks_single_region_tests.yml
+++ b/.github/workflows/aws_kubernetes_eks_single_region_tests.yml
@@ -842,7 +842,16 @@ jobs:
                   vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
 
             - name: 🏁 Install Camunda 8 using the generic/kubernetes helm chart procedure
-              timeout-minutes: 10
+              # Bumped from 10m: in domain mode install-chart.sh now also waits
+              # (bounded, fail-open) for the public Keycloak issuer to converge
+              # before returning, so the step needs room for the helm install plus
+              # that wait.
+              timeout-minutes: 22
+              env:
+                  # install-chart.sh waits for the public Keycloak issuer in domain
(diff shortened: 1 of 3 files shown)


leiicamundi:
Closing: redesigning per review feedback on #2718. The OIDC-issuer readiness wait will be **CI-only** (an internal composite action wired into the domain test workflows, not added to the customer-facing `install-chart.sh`) and **IdP-agnostic** (driven by the configured issuer, not Keycloak-specific). Replacement PRs to follow.

PR #2762, opened 2026-06-22
chore(ecs-dual-region): bump Camunda images to 8.10.0-alpha1

leiicamundi:
## What

The `8.10.0-alpha1` images are now published on Docker Hub, so the ECS dual-region `app` module no longer needs the `SNAPSHOT` tags (flagged by the alpha-availability check / InfEx Slack notification).

```
camunda/camunda:8.10-SNAPSHOT             -> camunda/camunda:8.10.0-alpha1
camunda/connectors-bundle:8.10-SNAPSHOT   -> camunda/connectors-bundle:8.10.0-alpha1
```

## Scope

These two `app/variables.tf` defaults were the **only** `org/image:<version>-SNAPSHOT` references in the repo (confirmed with the same pattern `internal_global_alpha_availability_check` uses, across `*.tf/*.sh/*.yml/*.yaml/*.tfvars`). `README.md` regenerated via terraform-docs.

The EC2 / Debian all-in-one `8.10.0-SNAPSHOT` references are **distribution tarballs**, not container images — they are not flagged by the check and intentionally stay on `SNAPSHOT` until 8.10 GA (see their `[release-duty]` TODOs).

## Verification

- `rg` for `org/image:<ver>-SNAPSHOT` under the module: none remaining.
- pre-commit: `terraform_fmt`, `tflint`, `terraform_docs` pass (trivy / terraform-test skipped locally — no Docker daemon / bash 3.2; run in CI).

### DIFF
--- a/aws/containers/ecs-dual-region-fargate/terraform/app/README.md
+++ b/aws/containers/ecs-dual-region-fargate/terraform/app/README.md
@@ -21,8 +21,8 @@
 | Name | Description | Type | Default | Required |
 | ---- | ----------- | ---- | ------- | :------: |
 | <a name="input_aws_profile"></a> [aws\_profile](#input\_aws\_profile) | AWS Profile to use (null = use default credential chain) | `string` | `null` | no |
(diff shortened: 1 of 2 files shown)


leiicamundi:
FYI @marcel-dias, I'm merging this one as it triggers an alert https://camunda.slack.com/archives/C076N4G1162/p1782119269522879

If it breaks something, feel free to revert, and then we can still exclude it in a followu pr

PR #2795, opened 2026-06-26
fix(ecs-dual-region): fix procedure scripts and document backend/teardown

hisImminence:
## Summary

Fixes bugs in the ECS dual-region procedure scripts validated by a full end-to-end deploy + failover on `eu-west-2` + `eu-west-3` (VPC Peering, RDBMS, `8.10-SNAPSHOT`).

## Issues Fixed

| # | Issue | Fix |
|---|-------|-----|
| 1 | `export_environment_prerequisites.sh` reads non-existent output names (`region_0_ecs_cluster_name`, `region_0_nlb_raft_endpoint`) → vars are `null` | Map to correct infra outputs: `ecs_cluster_region_0_id` (extract name from ARN), `nlb_raft_region_0_dns_name` |
| 2 | `verify_dual_region.sh` aborts at line 20 with cryptic bash error | Cascading from #1; also improved error message to explain how to fix |
| 3 | `verify_dual_region.sh` topology check returns 401 on 8.10 (no auth) → healthy cluster reported as unreachable | Add `-u "${ADMIN_USER}:${ADMIN_PASS}"` to all `/v2/*` curl calls |
| 4 | `failover.sh` planned path hangs ~10 min then times out | Root cause: after `failover-global-cluster`, queries for new writer ARN with `awk` but extracts the **old primary** identifier, then calls `wait_aurora_available` for that cluster in the wrong region. Fix: wait for global cluster to exit `switching-over` state, then wait on the known `SECONDARY_CLUSTER_ID` in `REGION_1`. |
| 5 | `failover.sh` Step 4 topology check also has no auth | Add `-u` flag |
| 6 | `failover.sh` switchover wait loop uses `local` outside a function → crash immediately after Aurora API call with `local: can only be used in a function` | Replace `local elapsed`, `local max_switchover_wait`, `local global_status` with plain assignments |
| 7 | `failback.sh` step 4 polls `/v2/topology` without auth → curl returns empty, quorum wait always times out at 600s | Add `-u "${ADMIN_USER}:${ADMIN_PASS}"` to the topology curl; require `ADMIN_USER`/`ADMIN_PASS` in prerequisites |
| 8 | No documentation for local backend usage | Added "Terraform backend" section with S3 and local examples |
| 9 | No guidance for teardown after failover | Added "Teardown after a failover" section with manual Aurora cleanup procedure |

## Validation

Full end-to-end deploy + failover on `eu-west-2` + `eu-west-3` (VPC Peering, RDBMS, local backend):

| Check | Result |
|-------|--------|
| `export_environment_prerequisites.sh` — all vars populated | ✅ PASS |
| `verify_dual_region.sh` — all 4 sections, 17 passed, 0 failed | ✅ PASS |
| Topology auth (8 brokers, RF=4, both regions) | ✅ PASS |
| `failover.sh` planned failover end-to-end, runtime 2m49s | ✅ PASS |
| Teardown (app 90 + infra 72 + vpc 52 resources) | ✅ PASS |

## Depends on
- #2277 (ECS dual-region reference architecture)


Szik:
@marcel-dias 
-  `failover.sh`  leaves cluster leaderless (only follower, error  503:  region0 down, aurora writer  switched to reagion1, tpology  fails with  0 leaders) 
-  `failback.sh`  afterwards hangs  (no  recover), looks  like  aurora  is read  only in region0  and  broker  in 0 crashloop  (aurora needsswitch back to  writer in region  0)

PR #2832, opened 2026-07-04
fix(ecs-dual-region): use single shared backup bucket across both regions

hisImminence:
## Summary

- Remove region 1 backup bucket and its IAM policy from `infra/s3.tf` — both regions now share region 0's bucket
- Remove orphaned `aws_kms_key.s3_region_1` and `aws_kms_alias.s3_region_1` from `infra/kms.tf` (sole consumer was the deleted bucket)
- Remove stale `backup_bucket_region_1_name` and `s3_backup_access_policy_region_1_arn` outputs from `infra/outputs.tf`
- Update region 1 module in `app/camunda.tf` to reference region 0's bucket name and IAM policy
- Add explicit `CAMUNDA_DATA_BACKUP_S3_REGION` env var on both modules (set to region 0, the bucket's home region)

## Why

Confirmed finding by Andreas Hauser (tested and passing), confirmed by Carlo Sana (2026-07-03): with per-region buckets, restore fails because Zeebe partition leaders can be in either region, so all backup data lands in one bucket while the other stays empty. A single shared bucket ensures all backup data is in one place regardless of which region is active.

## Migration note for existing deployments

Removing the region 1 backup bucket resource will cause `terraform apply` to destroy it. If the bucket contains existing backup data, either empty it manually first or set `s3_force_destroy = true` in `terraform/infra/terraform.tfvars` before applying.

## Test plan

- [ ] `terraform plan` on `infra/` shows destruction of region 1 backup bucket, its IAM policy, and its KMS key/alias only — no other unexpected changes
- [ ] `terraform plan` on `app/` shows updated env vars on region 1 module (`CAMUNDA_DATA_BACKUP_S3_BUCKETNAME`, `CAMUNDA_DATA_BACKUP_REPOSITORYNAME`, `CAMUNDA_DATA_BACKUP_S3_REGION`) and updated `extra_task_role_attachments`
- [ ] CI integration test passes

### DIFF
--- a/aws/containers/ecs-dual-region-fargate/terraform/app/camunda.tf
+++ b/aws/containers/ecs-dual-region-fargate/terraform/app/camunda.tf
@@ -54,6 +54,10 @@ module "orchestration_cluster_region_0" {
         name  = "CAMUNDA_DATA_BACKUP_REPOSITORYNAME"
         value = local.infra.backup_bucket_region_0_name
       },
+      {
+        name  = "CAMUNDA_DATA_BACKUP_S3_REGION"
+        value = data.aws_region.region_0.id
+      },
     ]
   )
 
@@ -148,11 +152,18 @@ module "orchestration_cluster_region_1" {
     [
       {
         name  = "CAMUNDA_DATA_BACKUP_S3_BUCKETNAME"
-        value = local.infra.backup_bucket_region_1_name
+        value = local.infra.backup_bucket_region_0_name
       },
       {
         name  = "CAMUNDA_DATA_BACKUP_REPOSITORYNAME"
-        value = local.infra.backup_bucket_region_1_name
+        value = local.infra.backup_bucket_region_0_name
+      },
+      {
+        # Explicit bucket region: region 1 brokers write cross-region to the
+        # shared bucket in region 0; the S3 client must know where the bucket
+        # lives for correct endpoint resolution.
+        name  = "CAMUNDA_DATA_BACKUP_S3_REGION"
+        value = data.aws_region.region_0.id
       },
     ]
   )
@@ -177,7 +188,7 @@ module "orchestration_cluster_region_1" {
 
   extra_task_role_attachments = concat(
     local.infra.rds_db_connect_policy_region_1_arn != null ? [local.infra.rds_db_connect_policy_region_1_arn] : [],
-    [local.infra.s3_backup_access_policy_region_1_arn],
+    [local.infra.s3_backup_access_policy_region_0_arn],
   )
 
   # See region 0 comments above.
--- a/aws/containers/ecs-dual-region-fargate/terraform/infra/README.md
+++ b/aws/containers/ecs-dual-region-fargate/terraform/infra/README.md
@@ -21,7 +21,6 @@
 | [aws_iam_policy.rds_db_connect_region_0](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy) | resource |
 | [aws_iam_policy.rds_db_connect_region_1](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy) | resource |
 | [aws_iam_policy.s3_backup_access_region_0](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy) | resource |
-| [aws_iam_policy.s3_backup_access_region_1](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy) | resource |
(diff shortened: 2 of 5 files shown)


leiicamundi:
## Architectural review — single shared backup bucket

I checked this change against the Camunda dual-region documentation and the existing EKS dual-region reference. Short version: the single shared bucket is the correct model and is consistent with the documented pattern. The notes below are strictly architectural.

### What I verified

**1. Runtime HA does not use this bucket.** The failover/failback path in this reference is Zeebe Raft quorum + Aurora Global Database promotion, not S3:
- `procedure/failover.sh` — scale down region 0, promote the Aurora Global secondary, verify Raft quorum in region 1.
- `procedure/failback.sh` — re-attach the Aurora secondary, scale region 0 back up, wait for cross-region Raft quorum.

Neither script references S3/backup/restore. This matches the concept docs: Zeebe "stretches across regions using the Raft protocol, allowing communication and data replication between all brokers" (`concepts/multi-region/dual-region.md:112`); on region loss "the Zeebe cluster loses quorum … and immediately stops processing new data" (`:254`), and recovery removes the lost brokers with "no data has been lost thanks to Zeebe data replication" (`operational-tasks/dual-region-ops.md:172`). So `CAMUNDA_DATA_BACKUP_S3_*` configures the **backup / disaster-recovery store**, not the HA mechanism.

**2. The documented dual-region pattern uses one cross-region bucket.** The operational procedure ("Create and restore Elasticsearch backup", `dual-region-ops.md:1161+`) registers a single snapshot repository against one bucket — `terraform output -raw s3_bucket_name` (singular) — from the surviving region, and restores it into the recreated region. The backup is created fresh from the survivor at failback time; both regions use the same bucket cross-region.

**3. The canonical EKS reference already uses one shared bucket.** `aws/kubernetes/eks-dual-region/terraform/clusters/s3.tf` defines a single `aws_s3_bucket.elastic_backup` (no per-region provider alias), one KMS key, one IAM policy, one `s3_bucket_name` output, accessed cross-region by both clusters. This PR brings the ECS reference in line with that established pattern.

### Architectural opinion

- **The change is correct.** Per-region buckets are effectively broken for restore in an active-active dual-region cluster: partition leaders can sit in either region, so each region's leaders write to their local bucket and no single bucket ever holds a complete backup. A single shared bucket is the only model that produces a restorable backup, and it is the pattern both the docs and the EKS reference already assume. Setting `CAMUNDA_DATA_BACKUP_S3_REGION` to the bucket's home region is required so the region-1 client resolves the correct cross-region endpoint (otherwise the SDK infers region 1 and misroutes).

- **"Isn't this the HA bucket?" — no.** HA/failover is Raft (+ Aurora Global DB). This is the backup/DR store, so consolidating it does not weaken the failover path.

- **One by-design limitation worth a doc note (not a blocker):** the shared bucket lives in a single region and therefore inherits that region's availability. While the bucket's home region (region 0) is down, the surviving region cannot take new backups, since the store points at region 0. The documented failback tolerates this because it creates a fresh backup from the survivor *after* the lost region — and its bucket — has been recreated. If a stronger DR posture is wanted (backup durable to the loss of *either* cluster region), the backup store should be independent of both regions: a bucket in a third region, an S3 Multi-Region Access Point, or CRR to a secondary bucket. The EKS reference has the same limitation today, so keeping it out of scope here is reasonable — I would just add a short README note about the single-region durability of the backup store.

Net: architecturally sound and consistent with the documented dual-region model; the only follow-up I'd suggest is documenting the backup store's single-region durability.

leiicamundi:
@hisImminence @marcel-dias — flagging one point from the review above for your call. Architectural only, **not a blocker** for this PR:

> **One by-design limitation worth a doc note:** the shared bucket lives in a single region and inherits that region's availability. While region 0 is down, the surviving region cannot take new backups, since the store points at region 0. The documented failback tolerates this because it creates a fresh backup from the survivor *after* the lost region — and its bucket — has been recreated. For a stronger DR posture (backup durable to the loss of *either* cluster region), the backup store could be made independent of both regions: a bucket in a third region, an S3 Multi-Region Access Point, or CRR to a secondary bucket.

The EKS dual-region reference has the same single-region-bucket limitation today, so I'd keep it out of scope here and just add a short README note about the backup store's single-region durability. Does that match your intent, or do you want a follow-up to harden it?

marcel-dias (approved):

marcel-dias:
@leiicamundi the idea is to ship as is and evaluate a multi-region HA setup for the S3 buckets in near future 
https://github.com/camunda/team-infrastructure-experience/issues/1173

hisImminence:
Perfect! Thank you team for moving this! 

> The EKS dual-region reference has the same single-region-bucket limitation today, so I'd keep it out of scope here 
[We are aligned](https://camunda.slack.com/archives/C0AU6ML2VGD/p1783358649293039?thread_ts=1782808028.017719&cid=C0AU6ML2VGD) ;)

PR #2867, opened 2026-07-08
Small changes during Dave's first install

upgradingdave:


aws/containers/ecs-dual-region-fargate/procedure/export_environment_prerequisites.sh
```diff
@@ -10,20 +10,19 @@
 # Must be sourced (not executed) to export variables to current shell.        #
 ###############################################################################
 
-set -euo pipefail
```
leiicamundi, line 13: Removing `set -euo pipefail` here is half-right, and worth being deliberate about since this file is **sourced** (`. ./export_environment_prerequisites.sh`):

- `set -e` in a sourced script leaks into the caller's shell — a single failing command then exits that shell, i.e. an interactive terminal just closes. That's almost certainly what bit you on first install, so dropping `-e` is defensible.
- But deleting the whole line also removes failure detection, which we *do* want to keep. The critical call below —
  ```bash
  TF_OUTPUT=$(terraform -chdir="${TF_DIR}" output -json)
  ```
  now fails silently: if terraform isn't initialized / wrong dir / no state, `TF_OUTPUT` is empty, every `jq` extraction returns empty, the summary prints blanks, and empty vars still get `export`ed. `verify_dual_region.sh` then blows up later with confusing `:?` errors.

Suggested middle ground — keep `pipefail` (harmless when sourced) and guard the one call that matters instead of relying on `set -e`:
```bash
set -o pipefail
...
if ! TF_OUTPUT=$(terraform -chdir="${TF_DIR}" output -json); then
    echo "ERROR: terraform output failed in ${TF_DIR}" >&2
    return 1 2>/dev/null || exit 1
fi
```

aws/containers/ecs-dual-region-fargate/procedure/export_environment_prerequisites.sh
```diff
@@ -10,20 +10,19 @@
...
 
-# Path to Terraform root (relative to this script). With the 3-state layout
-# (vpc/, infra/, app/), most outputs needed for environment setup live in
-# infra/. Override TF_DIR if using an alternate state location.
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-TF_DIR="${TF_DIR:-${SCRIPT_DIR}/../terraform/infra}"
+# Path to Terraform root — anchored to the git repo root so this script works
+# regardless of which directory it is sourced from.
+# Override TF_DIR to point at an alternate state location.
+_GIT_ROOT="$(git rev-parse --show-toplevel 2>/dev/null)"
```
leiicamundi, line 23: **please change** this regresses portability versus the old `BASH_SOURCE` approach. Outside a git checkout (release tarball, a copied procedure dir) `git rev-parse --show-toplevel` prints nothing (error silenced by `2>/dev/null`), so `TF_DIR` becomes `/aws/containers/.../terraform/infra` — absolute from `/` — and terraform fails. It also hardcodes the full repo-relative path, so moving this file silently breaks it.

The previous `SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` was already independent of the caller's CWD (the stated goal) *and* worked without git. Recommend keeping it:
```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
TF_DIR="${TF_DIR:-${SCRIPT_DIR}/../terraform/infra}"
```

aws/containers/ecs-dual-region-fargate/procedure/export_environment_prerequisites.sh
```diff
@@ -33,18 +32,24 @@ echo "Fetching Terraform outputs from ${TF_DIR} ..."
 
 TF_OUTPUT=$(terraform -chdir="${TF_DIR}" output -json)
 
+# Override region defaults from Terraform outputs if available
+_TF_REGION_0=$(echo "${TF_OUTPUT}" | jq -r '.region_0.value // empty')
+_TF_REGION_1=$(echo "${TF_OUTPUT}" | jq -r '.region_1.value // empty')
```
leiicamundi, line 37: nit: good change to sync `REGION_0`/`REGION_1` with what infra actually deployed. For consistency with the `unset _GIT_ROOT` above, also `unset _TF_REGION_0 _TF_REGION_1` after use — otherwise these two temporaries leak into the sourced interactive shell.

leiicamundi (commented):
Reviewed both procedure scripts.

**Correct changes** (verified against Terraform): the output-name fixes are right — `ecs_cluster_region_*_id` + `split("/") | last`, `nlb_raft_region_*_dns_name`, and `region_*` all exist in `terraform/infra/outputs.tf`, whereas the old `region_0_ecs_cluster_name` / `region_0_nlb_raft_endpoint` names never existed, so this fixes real breakage. The basic-auth addition is also correct: `${CLUSTER_0%-cluster}-oc-admin-user-password` matches the secret name in `secrets.tf`, the stored value is a raw password, and both regions share the same `random_password.admin_user_password`, so fetching region 0's secret and reusing it for region 1 is fine.

**Main point** (on the pipefail instinct): dropping `set -e` from a *sourced* script is actually the right call, but removing the whole line also drops error detection on the `terraform output` call — details inline. A few smaller notes inline too.

aws/containers/ecs-dual-region-fargate/procedure/verify_dual_region.sh
```diff
@@ -120,7 +133,7 @@ check_topology() {
     local label=$2
 
     local topology
-    topology=$(curl -sf --max-time 15 "http://${alb}/v2/topology" 2>/dev/null || echo "")
+    topology=$(curl -sf --max-time 15 -u "${ADMIN_USER}:${ADMIN_PASS}" "http://${alb}/v2/topology" 2>/dev/null || echo "")
```
leiicamundi, line 136: nit: `-u "${ADMIN_USER}:${ADMIN_PASS}"` puts the password in the process argv, visible via `ps` / `/proc` to other local users (same for the POST call at ~L226). Acceptable for a demo, but `--netrc-file` or an explicit `-H "Authorization: Basic ..."` header avoids leaking it.

aws/containers/ecs-dual-region-fargate/procedure/export_environment_prerequisites.sh
leiicamundi: Concrete **Option A** (execute + `eval`) — full `set -euo pipefail`, runs in its own process so a failure can never touch (or close) your shell.

Validated locally: `shellcheck` clean, and smoke-tested both paths — success emits `export …` lines on stdout + the human summary on stderr; a `terraform` failure exits non-zero with **empty stdout**, so `eval` becomes a no-op and the terminal survives with no half-populated env.

```bash
#!/bin/bash

###############################################################################
# Export environment variables for ECS Dual-Region Fargate procedures         #
#                                                                             #
# Usage:                                                                      #
#   eval "$(./export_environment_prerequisites.sh)"                           #
#                                                                             #
# Runs in its OWN process under `set -euo pipefail`. Any failure aborts with  #
# a non-zero exit and prints nothing on stdout, so a broken run never leaves  #
# a half-populated environment and never affects (or closes) your shell.      #
# Diagnostics/summary go to stderr; only `export ...` lines go to stdout for  #
# `eval` to consume.                                                          #
#                                                                             #
# Requires: terraform CLI, awscli, jq                                         #
###############################################################################

set -euo pipefail

###############################################################################
# User-configurable defaults                                                  #
###############################################################################

REGION_0="${REGION_0:-eu-west-2}"
REGION_1="${REGION_1:-eu-west-3}"

# Path to Terraform root — relative to this script, independent of the caller's
# CWD (works whether run from the procedure dir or elsewhere, git or not).
# Override TF_DIR to point at an alternate state location.
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
TF_DIR="${TF_DIR:-${SCRIPT_DIR}/../terraform/infra}"

###############################################################################
# Retrieve Terraform outputs                                                  #
###############################################################################

echo "Fetching Terraform outputs from ${TF_DIR} ..." >&2

TF_OUTPUT="$(terraform -chdir="${TF_DIR}" output -json)"

# Override region defaults from Terraform outputs when present
_TF_REGION_0="$(jq -r '.region_0.value // empty' <<<"${TF_OUTPUT}")"
_TF_REGION_1="$(jq -r '.region_1.value // empty' <<<"${TF_OUTPUT}")"
[ -n "${_TF_REGION_0}" ] && REGION_0="${_TF_REGION_0}"
[ -n "${_TF_REGION_1}" ] && REGION_1="${_TF_REGION_1}"

# Region 0
CLUSTER_0="$(jq -r '(.ecs_cluster_region_0_id.value // "") | split("/") | last' <<<"${TF_OUTPUT}")"
ALB_ENDPOINT_0="$(jq -r '.region_0_alb_endpoint.value' <<<"${TF_OUTPUT}")"
NLB_GRPC_ENDPOINT_0="$(jq -r '.region_0_nlb_grpc_endpoint.value' <<<"${TF_OUTPUT}")"
NLB_RAFT_ENDPOINT_0="$(jq -r '.nlb_raft_region_0_dns_name.value' <<<"${TF_OUTPUT}")"

# Region 1
CLUSTER_1="$(jq -r '(.ecs_cluster_region_1_id.value // "") | split("/") | last' <<<"${TF_OUTPUT}")"
ALB_ENDPOINT_1="$(jq -r '.region_1_alb_endpoint.value' <<<"${TF_OUTPUT}")"
NLB_GRPC_ENDPOINT_1="$(jq -r '.region_1_nlb_grpc_endpoint.value' <<<"${TF_OUTPUT}")"
NLB_RAFT_ENDPOINT_1="$(jq -r '.nlb_raft_region_1_dns_name.value' <<<"${TF_OUTPUT}")"

# Aurora Global Database
AURORA_GLOBAL_CLUSTER_ID="$(jq -r '.aurora_global_cluster_id.value' <<<"${TF_OUTPUT}")"
AURORA_PRIMARY_ENDPOINT="$(jq -r '.aurora_primary_endpoint.value' <<<"${TF_OUTPUT}")"
AURORA_SECONDARY_ENDPOINT="$(jq -r '.aurora_secondary_endpoint.value' <<<"${TF_OUTPUT}")"

###############################################################################
# Summary (stderr — keeps stdout clean for eval)                              #
###############################################################################

cat >&2 <<EOF

=== ECS Dual-Region Environment ===

Region 0 (${REGION_0}):
  ECS Cluster:    ${CLUSTER_0}
  ALB (HTTP):     ${ALB_ENDPOINT_0}
  NLB (gRPC):     ${NLB_GRPC_ENDPOINT_0}
  NLB (Raft):     ${NLB_RAFT_ENDPOINT_0}

Region 1 (${REGION_1}):
  ECS Cluster:    ${CLUSTER_1}
  ALB (HTTP):     ${ALB_ENDPOINT_1}
  NLB (gRPC):     ${NLB_GRPC_ENDPOINT_1}
  NLB (Raft):     ${NLB_RAFT_ENDPOINT_1}

Aurora Global Database:
  Global Cluster: ${AURORA_GLOBAL_CLUSTER_ID}
  Primary (W):    ${AURORA_PRIMARY_ENDPOINT}
  Secondary (R):  ${AURORA_SECONDARY_ENDPOINT}

EOF

###############################################################################
# Emit exports on stdout ONLY. Reached only on full success, so `eval` never  #
# sees a partial environment. printf %q keeps values safe to eval.            #
###############################################################################

for _var in REGION_0 REGION_1 \
            CLUSTER_0 ALB_ENDPOINT_0 NLB_GRPC_ENDPOINT_0 NLB_RAFT_ENDPOINT_0 \
            CLUSTER_1 ALB_ENDPOINT_1 NLB_GRPC_ENDPOINT_1 NLB_RAFT_ENDPOINT_1 \
            AURORA_GLOBAL_CLUSTER_ID AURORA_PRIMARY_ENDPOINT AURORA_SECONDARY_ENDPOINT; do
    printf 'export %s=%q\n' "${_var}" "${!_var}"
done
```

**Invocation contract changes** (`source` → `eval "$(…)"`), so the callers/doc need a one-line update:

- `README.md` (step 4): `source ../../procedure/export_environment_prerequisites.sh` → `eval "$(../../procedure/export_environment_prerequisites.sh)"`
- header usage note in this file and the siblings that mention `. ./export_environment_prerequisites.sh` — `verify_dual_region.sh`, `failover.sh`, `failback.sh`, `test_cross_region_dns.sh`

The consumers keep working unchanged: `eval` still `export`s the vars into the current shell, so their `${REGION_0:?…}` guards are satisfied exactly as before.

PR #2911, opened 2026-07-20
test(ecs-dual-region): add Terraform golden plan coverage for vpc, infra and app

marcel-dias:
## Motivation

The ECS dual-region Fargate reference architecture (`aws/containers/ecs-dual-region-fargate/`) ships three Terraform states — `vpc`, `infra`, `app` — but none had golden plan coverage. `just regenerate-golden-file-all` (run by the scheduled maintenance job) expects a golden setup per state, and the gap previously broke that job until #2761 made the recipe skip states without `test/golden/golden.tfvars`. As a result these three states were the only reference-architecture modules excluded from golden plan drift detection.

Closes the coverage gap tracked in [team-infrastructure-experience#1158](https://github.com/camunda/team-infrastructure-experience/issues/1158).

## What this does

Adds golden plan coverage for **all three** states, plus a CI workflow that compares each on PRs.

| State | Plannable standalone? | Approach |
|-------|----------------------|----------|
| `vpc` | Yes (only `data.aws_availability_zones`) | `golden.tfvars` + generated `tfplan-golden.json` |
| `infra` | No — reads `terraform_remote_state.vpc` + live `data.aws_subnet` | fixture override + generated golden |
| `app` | No — reads `terraform_remote_state.infra` | fixture override + generated golden |

`infra` and `app` cannot be planned standalone because they read a sibling state over an S3 backend that does not exist during golden generation. Each ships a `test/fixtures/golden/fixture_*_override.tf` (a Terraform override file the `regenerate-golden-file` recipe copies in and removes) that:

- disables the remote-state read (`count = 0`) so no backend is contacted, and
- injects a static, deterministic snapshot of the upstream outputs.

`infra` additionally stubs the live `data.aws_subnet` lookups and re-injects the derived AZ list. `data.aws_caller_identity` stays live — the account ID it yields appears only inside ARNs, which the recipe redacts.

`app` is the most valuable of the three to cover: its golden captures the Camunda ECS task definitions, services and environment variables.

## Implications

- All three states now provide `test/golden/golden.tfvars`, so they participate in `regenerate-golden-file-all` and golden drift detection like every other reference architecture.
- New workflow `.github/workflows/aws_ecs_dual_region_fargate_golden.yml` runs the three comparisons on PRs. Its `paths:` triggers cover the reference architecture's terraform, `.tool-versions`, and the shared modules rendered into the golden plans (`ecs`, `aurora-global`, `opensearch`, `transit-gateway`) so a change to any of them re-runs the comparison.
- The `regenerate-golden-file` recipe now removes copied fixtures via an `EXIT` trap, so a failed run cannot leave a stray `fixture_*_override.tf` in a module root (which would silently neutralize the real remote-state read on a subsequent standalone plan).

## Verification

- All three golden files were generated against the CI state backend, are **deterministic** (fresh regeneration matches the committed file via `delta`), and are **redaction-clean** (no real account IDs, ARNs, developer paths, or emails).
- `pre-commit run --all-files`-equivalent hooks pass (terraform fmt/tflint/docs, trivy, yamllint, actionlint, and the `terraform test` suites for the touched states).

## Notes

Rebased on top of #2910, which repaired the `infra`/`app` `terraform test` suites and renamed `aurora_secondary_endpoint` → `aurora_secondary_cluster_endpoint`; the `app` fixture uses the corrected output name.

### DIFF
--- a/.github/workflows/aws_ecs_dual_region_fargate_golden.yml
+++ b/.github/workflows/aws_ecs_dual_region_fargate_golden.yml
@@ -1,0 +1,105 @@
+---
+name: Tests - Golden - AWS Containers ECS Dual Region Fargate
+
+on:
+    workflow_dispatch:
+    pull_request:
+        paths:
+            - .github/workflows/aws_ecs_dual_region_fargate_golden.yml
+            - .github/actions/aws-configure-cli/**
+            - .github/actions/internal-terraform-golden-plan/**
+            - .tool-versions
+            - aws/containers/ecs-dual-region-fargate/terraform/**
+            - aws/modules/transit-gateway/**
+
+# limit to a single execution per actor of this workflow
+concurrency:
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+env:
+    IS_SCHEDULE: ${{ (contains(github.head_ref, 'schedules/') || github.event_name == 'schedule') && 'true' || 'false' }}
+
+    AWS_PROFILE: infraex
+    AWS_REGION: eu-west-2
+    S3_BACKEND_BUCKET: tests-ra-aws-rosa-hcp-tf-state-eu-central-1
+    S3_BUCKET_REGION: eu-central-1
+    S3_BUCKET_KEY: golden.tfstate
+
+    MODULE_DIR_BASE: ./aws/containers/ecs-dual-region-fargate/terraform/
+
+jobs:
+    triage:
+        runs-on: ubuntu-slim
+        outputs:
+            should_skip: ${{ steps.skip_check.outputs.should_skip }}
+        steps:
+            - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+            - name: Check labels
+              id: skip_check
+              uses: ./.github/actions/internal-triage-skip
+
+    compare:
+        runs-on: ubuntu-latest
+        needs:
+            - triage
+        if: needs.triage.outputs.should_skip == 'false'
+        permissions:
+            contents: write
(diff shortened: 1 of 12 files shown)


hisImminence (approved):
**PR #2911: test(ecs-dual-region): add Terraform golden plan coverage for vpc, infra, app**

Checked workflow, justfile diff, both fixture overrides, golden.tfvars for all 3 states, and cross-referenced sibling golden workflows (`aws_ecs_single_region_fargate_golden.yml`, `aws_kubernetes_eks_dual_region_golden.yml`) + module `variables.tf` for required-var gaps. No blockers found.

- `justfile:34`: 🔵 nit: `trap 'rm -f fixture_*.tf' EXIT` now leaves fixture copies present through `terraform show -json`/redaction/jq steps instead of removing right after `plan`. Harmless (no later step touches them) but worth a one-line comment noting it's intentional, not an oversight.
- `fixture_infra_override.tf:L33-34`: 🔵 nit: comment explains why `region_1_azs` is deliberately omitted — good, keep this pattern for future fixture additions.
- app `variables.tf:61,71`: ❓ q: confirmed `camunda_image`/`connectors_image` have SNAPSHOT defaults so golden.tfvars doesn't need them — no action, just flagging I checked this since golden.tfvars looked thin at first glance.
- `aws_ecs_dual_region_fargate_golden.yml`: matches sibling workflow conventions (`ubuntu-slim` triage, shared `S3_BACKEND_BUCKET` name, matrix over module dirs) — no drift from established pattern.

Overall: solid, well-documented fixture pattern (`count = 0` + static snapshot) for breaking cross-state `terraform_remote_state` reads in golden generation. Redaction verified clean. Good to merge pending CI green.

PR #2917, opened 2026-07-21
feat(ecs): Management Identity, Keycloak, and OIDC authentication mode for single-region ECS Fargate

marcel-dias:
Resolves camunda/team-infrastructure-experience#1181

## Summary

Extends the **AWS single-region ECS Fargate** reference architecture (`aws/containers/ecs-single-region-fargate`) with the Camunda **Management Identity** component, a bundled **Keycloak** identity provider, and a pluggable **`authentication_mode`** (`basic` / bundled `keycloak` / `external` OIDC) for the whole platform. Prior to this, the deployment ran the Orchestration Cluster + Connectors on basic auth with no IdP.

Delivered incrementally and validated end-to-end on a live throwaway cluster (`eu-west-3`). Implements issue camunda/team-infrastructure-experience#1181 (PDP camunda/product-hub#3432, target 8.10) and its natural follow-ups.

> **Not production-ready.** Like every reference architecture here, this is a demo/learning blueprint. See _Known limitations_ — notably the OIDC-over-HTTP posture.

---

### 1. Management Identity (module `aws/modules/ecs/fargate/management-identity`)
- New ECS Fargate module modeled on the `connectors` sibling; image `camunda/identity:8.10.0-alpha3` (Renovate-tracked).
- Dedicated `identity` database + **password** auth on the shared Aurora cluster (Identity's image has no AWS IAM JDBC wrapper — deliberate divergence from the orchestration cluster's IAM auth).
- Exposed internally via Service Connect (`identity:8084`); liveness health check on the management port; ALB exposure opt-in (default off).
- Idempotent DB seed (guarded `CREATE DATABASE` without `OWNER` for RDS compatibility).

### 2. Keycloak + realm bootstrap (module `aws/modules/ecs/fargate/keycloak`)
- New ECS Fargate module running `camunda/keycloak:quay-26.6.4` (`start`), ports **18080** (`http`, Service Connect `keycloak`) + **9000** (`management`), `/dev/tcp` bash health probe on `/auth/health/ready`.
- Dedicated `keycloak` Aurora database (password auth), admin + DB password secrets in Secrets Manager (CMK-encrypted).
- Management Identity bootstraps the **`camunda-platform` realm** on startup from a single `SPRING_APPLICATION_JSON` (assembled in the cluster root as `identity_realm.tf`). Client secrets are Secrets Manager values injected as env vars and referenced as `${...}` placeholders — **never inlined** in the JSON.
- Per-component OIDC client flags (`enable_*_oidc_client`): `orchestration`/`connectors` default on, `optimize`/`console`/`web-modeler` default off; `identity` always. Each flag gates the realm preset, the client secret, and the injected env together. (web-modeler is a public client → no secret.)

### 3. `authentication_mode` toggle (`auth_mode.tf`) — **new capability**
A single flag `authentication_mode = basic | keycloak | external` (default `basic`). All provider-dependent values (issuer, token URL, audience, per-component client IDs + secret ARNs) resolve **once** into `local.oidc_*`, so the module blocks never branch on the provider.

- **`basic`** (default) — Orchestration/Connectors use built-in users. The bundled Keycloak + Management Identity are still deployed (Identity needs an IdP), so the reference stays self-contained and can flip to `keycloak` without adding infrastructure.
- **`keycloak`** — full OIDC against the bundled Keycloak `camunda-platform` realm that Identity bootstraps: Orchestration on the OIDC security method (audience `orchestration-api`, `preferred_username` claim), Connectors on **client-credentials** (m2m), Keycloak exposed on the shared ALB (`/auth*`) for the browser login redirect, and the ALB URL as the token issuer so `iss` is identical for browser, orchestration discovery, and connectors. The realm login user is **`admin`** with a **generated 32-char password** in Secrets Manager (replaces the previous `demo/demo`).
- **`external`** — **bring-your-own OIDC (Entra ID / Okta): Keycloak is skipped entirely** (its module, DB seed, secrets, and `/auth*` ALB rule are all `count`-gated off). Orchestration, Connectors and Management Identity point at a single typed `external_oidc` object — issuer/token URIs, audience, and one client id + **Secrets-Manager secret ARN** per component (identity, orchestration, connectors). A `terraform_data` precondition fails fast if `external` is selected without it; client secrets are passed as existing ARNs, never raw values.

> **`external` is structurally complete but not yet live-validated** (needs a real Entra/Okta tenant). Management Identity's generic-OIDC property keys are marked to validate on first deploy. `basic` and `keycloak` are the live-validated paths.

---

## Auth flows (`keycloak` mode)

```
Browser → ALB (/*) → Orchestration → 302 → ALB (/auth*) → Keycloak login
       → back to ALB /sso-callback → Orchestration (authorization_code + PKCE)

Connectors → ALB (/auth*) → Keycloak token endpoint (client_credentials)
          → Bearer token (aud: orchestration-api) → Orchestration /v2 API

Management Identity → keycloak:18080 (internal Service Connect) → admin API
          → creates realm + clients + admin user
```

In **`external`** mode this is replaced by the customer's IdP: the same flows point at `external_oidc.issuer_uri`/`token_uri`, no Keycloak is deployed, and Identity runs in generic-OIDC mode (no realm bootstrap — the external IdP owns clients/users).

---

## Live validation (throwaway cluster, eu-west-3)

Deployed from local Terraform and verified end-to-end:

- **Keycloak**: schema migration, server start (18080/9000), admin bootstrap, container HEALTHY.
- **Management Identity**: connects to Keycloak, bootstraps the `camunda-platform` realm, reaches HEALTHY (resolves the pre-IdP crash-loop). Verified via the Keycloak admin API: realm present, clients `camunda-identity`/`orchestration`/`connectors` (+ `-api`/`-resource-server`) created, disabled components correctly absent, `admin` user created with the platform roles.
- **OIDC end-to-end**:
  - **Operate login** — `/operate` → Keycloak → `admin` / generated password → admin session.
  - **Token issuance** — decoded a realm token (`aud: [connectors, orchestration-api, account]`, `iss: <ALB>/auth/realms/camunda-platform`).
  - **Deploy** — deployed a BPMN (`bigVarProcess`) via the OIDC-authenticated `POST /v2/deployments`.
  - **Runtime** — created multiple `bigVarProcess` instances via `POST /v2/process-instances`.
- **Connectors**: switches to OIDC client-credentials and authenticates to the OIDC orchestration cluster.

---

## Fixes discovered during live validation (each its own commit)

These are things plan-time/`terraform validate` cannot catch — they only surface against real ECS/Keycloak/Camunda:

| Fix | Why |
|---|---|
| `concat()` instead of `compact()` for `ecs_task_secret_arns` | `compact()` over "known after apply" ARNs makes the IAM policy `count = length(...)>0` unknown at plan time → `Invalid count argument`. |
| Keycloak health-check `retries` 15 → 10 | ECS caps health-check retries at 10 (`RegisterTaskDefinition` 400). |
| Explicit `["CMD","bash","-c",...]` health probe | `/dev/tcp` is a bash-ism; `CMD-SHELL` (`/bin/sh`) isn't guaranteed to be bash. |
| `KC_CACHE=local` (single-instance) | Clustered Keycloak forms JGroups over 7800/57800 (not opened in the SG); a rolling redeploy's second node's cluster health check stays DOWN → circuit breaker. `local` cache avoids clustering for `task_desired_count=1`. |
| Keycloak/Identity ALB rule priority below 100 | The orchestration catch-all `/*` at priority 100 shadowed the `/auth*` (160) and `/identity*` (150) rules; moved to 40/30. |
| OIDC audience `orchestration-api` | The realm issues tokens with `aud` including `orchestration-api`, not `orchestration`. |

---

## Known limitations / operational notes (important for docs)

- **`keycloak` mode over HTTP requires `sslRequired=NONE` on the realm.** With the ALB on HTTP (no domain/cert in this ref-arch), Keycloak's default realm `sslRequired=external` rejects OIDC discovery with `HTTPS required`. During live validation this was set once via the Keycloak admin API. **This is not yet expressed in Terraform** — a fresh `authentication_mode=keycloak` deploy needs the realm `sslRequired` relaxed (or, preferably, TLS). Follow-up options: a durable IaC step (set `sslRequired` post-realm-creation), or front the ALB with **HTTPS (ACM cert + `KC_HOSTNAME` + `KC_PROXY_HEADERS=xforwarded`)** — the genuinely secure production path. (`external` mode avoids this entirely — an external IdP is already HTTPS.)
- **`basic` mode is the default and is unchanged** — the golden plan is generated with `authentication_mode=basic`; the OIDC env/clients only appear when the flag is set, and `external`-mode resources never appear in the default golden.
- The realm `admin` login password is in Secrets Manager (`${prefix}-oc1-realm-admin-user-password`); the old `demo` realm user may linger from earlier deploys but is unmapped/unused.
- Keycloak runs `start` (not `start --optimized`), so each boot re-augments the image (~30s) and logs a benign slf4j `AbstractMethodError` during the version sanity check (upstream image quirk, non-fatal).

---

## Testing

- Native Terraform tests for the new modules (`terraform test`, mock provider).
- Cluster golden plan regenerated (`test/golden/tfplan-golden.json`) with redaction verified (no account IDs / ARNs / emails / IPs).
- Live deploy + teardown against `eu-west-3`.

## Follow-ups

- **Live-validate `external` mode** against a real Entra ID / Okta tenant (the toggle is implemented; only `basic`/`keycloak` have been deployed and verified).
- Make the `keycloak`-mode `sslRequired`/HTTPS handling durable in IaC (see limitations).
- OIDC-aware demo script (Bearer variant of `demo-create-instances.sh`).
- Companion documentation in `camunda-docs` (this PR description is the source material).


aws/modules/ecs/fargate/management-identity/lb.tf
```diff
@@ -0,0 +1,58 @@
...
+    type            = "lb_cookie"
+    cookie_duration = 43200
+  }
+}
+
+// Reuse the shared ALB listener via path-based routing. Priority MUST be lower
+// than the orchestration catch-all rule (`/*` at priority 100), which is
+// evaluated first and would otherwise swallow every /identity* request. Uses 30,
+// unique vs connectors (50) and keycloak (40) and below the catch-all; the
+// path patterns are disjoint so their relative order does not matter.
```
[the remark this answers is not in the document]
marcel-dias: Good catch — corrected in 1ead2b4b. The comment now states ALB evaluates listener rules in ascending priority-number order (lowest number first), so this rule (30) must be *below* the orchestration catch-all (`/*` at 100) to be matched before it.

aws/modules/ecs/fargate/keycloak/lb.tf
```diff
@@ -0,0 +1,58 @@
...
+    type            = "lb_cookie"
+    cookie_duration = 43200
+  }
+}
+
+// Reuse the shared ALB listener via path-based routing. Priority MUST be lower
+// than the orchestration catch-all rule (`/*` at priority 100), which is
+// evaluated first and would otherwise swallow every /auth* request. Uses 40,
+// unique vs connectors (50) and below the catch-all; the /auth* and /connectors*
+// patterns are disjoint so their relative order does not matter.
```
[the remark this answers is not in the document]
marcel-dias: Fixed in 1ead2b4b — reworded to say ALB evaluates rules lowest-number-first, so this rule (40) must be below the orchestration catch-all (`/*` at 100) to match before it (otherwise `/*` swallows `/auth*`).

aws/containers/ecs-single-region-fargate/terraform/cluster/postgres_seed.tf
```diff
@@ -27,24 +27,66 @@ resource "aws_ecs_task_definition" "db_seed" {
...
 
-          for user in $${IAM_DB_USERS}; do
-            echo "Ensuring role exists: $${user}"
+            # Create/refresh the password-authenticated role
+            psql "host=$${AURORA_ENDPOINT} port=$${AURORA_PORT} dbname=$${AURORA_DB_NAME} user=$${AURORA_ADMIN_USERNAME} password=$${AURORA_ADMIN_PASSWORD} sslmode=require" \
+              -v ON_ERROR_STOP=1 \
+              -c "DO \$\$ BEGIN IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = '$${IDENTITY_DB_USERNAME}') THEN CREATE ROLE \"$${IDENTITY_DB_USERNAME}\" WITH LOGIN PASSWORD '$${IDENTITY_DB_PASSWORD}'; END IF; END \$\$;" \
+              -c "ALTER ROLE \"$${IDENTITY_DB_USERNAME}\" WITH LOGIN PASSWORD '$${IDENTITY_DB_PASSWORD}';"
+
+            # Create the dedicated database if it does not exist (owned by the identity role)
```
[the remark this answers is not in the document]
marcel-dias: Fixed in 1ead2b4b. The comment now correctly states no OWNER is set — the RDS master role cannot create a database owned by another role — and the identity role gets access via the GRANTs that follow.

aws/containers/ecs-single-region-fargate/terraform/cluster/variables.tf
```diff
@@ -101,6 +105,70 @@ variable "db_seed_iam_usernames" {
   default     = ["camunda"]
 }
 
+variable "identity_db_name" {
+  type        = string
+  description = "Dedicated database name for Management Identity on the shared Aurora cluster"
+  default     = "identity"
+}
```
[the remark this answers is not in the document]
marcel-dias: Added strict PostgreSQL-identifier validation in 1ead2b4b: `^[a-zA-Z_][a-zA-Z0-9_]*$` and length ≤ 63, so quotes/whitespace can't break the seed SQL or inject.

aws/containers/ecs-single-region-fargate/terraform/cluster/variables.tf
```diff
@@ -101,6 +105,70 @@ variable "db_seed_iam_usernames" {
...
+
+variable "identity_db_username" {
+  type        = string
+  description = "Password-authenticated database role for Management Identity (Identity does not support IAM DB auth)"
+  default     = "identity"
+}
```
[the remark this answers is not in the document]
marcel-dias: Added the same PostgreSQL-identifier validation (`^[a-zA-Z_][a-zA-Z0-9_]*$`, ≤ 63 chars) to identity_db_username in 1ead2b4b.

aws/containers/ecs-single-region-fargate/terraform/cluster/variables.tf
```diff
@@ -101,6 +105,70 @@ variable "db_seed_iam_usernames" {
...
+
+variable "keycloak_db_name" {
+  type        = string
+  description = "Dedicated database name for Keycloak on the shared Aurora cluster"
+  default     = "keycloak"
+}
```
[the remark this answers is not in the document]
marcel-dias: Added PostgreSQL-identifier validation (`^[a-zA-Z_][a-zA-Z0-9_]*$`, ≤ 63 chars) to keycloak_db_name in 1ead2b4b.

aws/containers/ecs-single-region-fargate/terraform/cluster/variables.tf
```diff
@@ -101,6 +105,70 @@ variable "db_seed_iam_usernames" {
...
+
+variable "keycloak_db_username" {
+  type        = string
+  description = "Password-authenticated database role for Keycloak"
+  default     = "keycloak"
+}
```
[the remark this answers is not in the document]
marcel-dias: Added PostgreSQL-identifier validation (`^[a-zA-Z_][a-zA-Z0-9_]*$`, ≤ 63 chars) to keycloak_db_username in 1ead2b4b.

hisImminence (changes requested):
## Review findings

**1. Major — bundled Keycloak OIDC login will not complete on a fresh deploy**

The shared ALB is HTTP-only with no domain/cert, so Keycloak's default realm `sslRequired=external` rejects OIDC discovery with "HTTPS required". During live validation this was worked around manually via the Keycloak admin API, but it is not expressed in Terraform (`auth_mode.tf`, `identity_realm.tf`). As written, a fresh `authentication_mode=keycloak` apply will not complete the browser login flow until someone manually relaxes `sslRequired` or fronts the ALB with TLS.

Since `keycloak` mode is one of the three primary paths this PR introduces, please either make this durable in IaC (a bootstrap step against the realm admin API, or wire up `KC_HOSTNAME` + ACM + `KC_PROXY_HEADERS=xforwarded`), or fail fast / call it out prominently in the module README so a user following the reference doesn't hit a silent dead end at login.

**2. `variables.tf:66-70` — `db_name` lacks the identifier validation applied to its new siblings**

`identity_db_name` / `identity_db_username` / `keycloak_db_name` / `keycloak_db_username` all received `^[a-zA-Z_][a-zA-Z0-9_]*$` + length ≤63 validation in this PR because they're interpolated into the same `postgres_seed.tf` SQL (quoted identifiers + `dbname=` conninfo string). `var.db_name` is interpolated identically (`AURORA_DB_NAME` → `dbname=$${AURORA_DB_NAME}`, `GRANT ALL PRIVILEGES ON DATABASE \"$${AURORA_DB_NAME}\"`) but never got the same guard. Since this PR establishes the validation pattern for exactly this class of variable, please extend it to `db_name` too for consistency.

**Nit:** `keycloak_admin_username` has no validation, unlike its DB-role siblings. Not an injection risk (never interpolated into SQL), but worth the same treatment if you're touching this block anyway.

**Note:** Copilot's prior findings (ALB rule priority comments, DB-owner comment, identifier validation) were already fixed and replied-to on-thread — no action needed there.

marcel-dias:
@hisImminence thanks for the review — all three points are addressed in `7bf60f64`, on top of the TLS listener already added in `c1edb3a6`.

**1 (Major) — `keycloak`-mode login over plain HTTP.**
The TLS path in `c1edb3a6` wires this durably in IaC: set `alb_certificate_arn` and the ALB gains an HTTPS `:443` listener, HTTP redirects to it, and Keycloak gets `KC_PROXY_HEADERS=xforwarded`. The ALB then forwards `X-Forwarded-Proto: https`, so the realm's default `sslRequired=external` is satisfied with **no manual step and no `sslRequired` relaxation**.

For the default HTTP demo I could not make `sslRequired=NONE` durable in IaC: the `camunda-platform` realm is provisioned **asynchronously by Management Identity at container startup** (via `SPRING_APPLICATION_JSON`), so it doesn't exist at Terraform apply time, and Keycloak exposes no realm-import/global hook to inject `sslRequired` without racing that bootstrap. So — per your "call it out prominently" option — I documented the limitation and both resolutions (enable TLS, or relax `sslRequired` via the admin API for an HTTP-only demo) prominently at the top of the deployment README (`aws/containers/ecs-single-region-fargate/README.md`). `basic` and `external` modes are unaffected (`basic` does no browser OIDC redirect against Keycloak; `external` uses the customer's own HTTPS provider).

**2 — `db_name` validation.**
Extended the same PostgreSQL-identifier validation (`^[a-zA-Z_][a-zA-Z0-9_]*$`, length ≤ 63) to `db_name`, since it's interpolated into the same `postgres_seed.tf` SQL as its siblings.

**Nit — `keycloak_admin_username`.**
Added the matching validation for consistency.

For reference, Copilot's earlier findings (ALB rule-priority comments, DB-owner comment, DB-identifier validation) were fixed in `1ead2b4b` and replied to on-thread — those threads are resolved.

aws/containers/ecs-single-region-fargate/terraform/cluster/auth_mode.tf
```diff
@@ -0,0 +1,123 @@
...
+
+locals {
+  oidc_enabled            = var.authentication_mode == "oidc"               # OIDC used at all
+  use_external            = local.oidc_enabled && var.external_oidc != null # bring-your-own IdP
+  deploy_bundled_keycloak = local.oidc_enabled && var.external_oidc == null # ship the default IdP
+
+  # Browser-facing base URL of the shared ALB. For the bundled Keycloak it is also the
+  # OIDC issuer host, so the token `iss` is identical for the browser and the backend
+  # (which reaches the ALB via NAT egress). HTTP only in this demo.
+  alb_base_url                = "http://${join("", aws_lb.main[*].dns_name)}"
```
leiicamundi: **Blocking — the TLS path cannot complete an OIDC flow.**

`alb_base_url` hardcodes `http://` and never reads `local.alb_https_enabled`. It feeds `keycloak_public_base_url` → `camunda_realm_issuer_public` → `local.oidc.issuer_uri`, `local.oidc.redirect_uri`, `identity_public_base`, and the realm `rootUrl` / `redirectUris` / `post.logout.redirect.uris`.

So when `alb_certificate_arn` is set:

1. `lb.tf:65-86` 301s every `:80` request to `:443`.
2. Keycloak, with `KC_PROXY_HEADERS=xforwarded` (`keycloak.tf:63`) and `hostname-strict=false`, derives its frontend URL from `X-Forwarded-Proto` and publishes `"issuer": "https://…/auth/realms/camunda-platform"`.
3. Camunda is configured with `CAMUNDA_SECURITY_AUTHENTICATION_OIDC_ISSUERURI = http://…`. Spring Security's `ClientRegistrations.fromIssuerLocation` requires the metadata `issuer` to be identical to the requested one, so it fails at startup.

Second problem in the same place: every public URL is built from `aws_lb.main.dns_name`, but ACM will not issue a public certificate for `*.elb.amazonaws.com`. Even with the scheme fixed, there is no way to supply the hostname the certificate actually covers.

Minimal fix: add an `alb_public_base_url` input (for example `https://camunda.example.com`), require it whenever `alb_certificate_arn` is set, derive every issuer / redirect / root URL from it, and pin `KC_HOSTNAME` to the same value.

This matters more than a normal latent bug because the TLS path is what was offered as the resolution to the earlier `sslRequired` Major finding.

aws/containers/ecs-single-region-fargate/terraform/cluster/auth_mode.tf
```diff
@@ -0,0 +1,123 @@
...
+  # CAMUNDA_IDENTITY_BASE_URL and for the camunda-identity client redirect in the
+  # bundled realm import).
+  identity_public_base = "${local.alb_base_url}/identity"
+
+  # Single provider-agnostic OIDC interface. Every component reads only this object;
+  # it is populated identically whether the IdP is the bundled Keycloak or external.
+  oidc = {
+    issuer_uri   = local.use_external ? try(var.external_oidc.issuer_uri, "") : local.camunda_realm_issuer_public
+    token_uri    = local.use_external ? try(var.external_oidc.token_uri, "") : "${local.camunda_realm_issuer_public}/protocol/openid-connect/token"
+    audience     = local.use_external ? try(var.external_oidc.audience, "") : "orchestration-api"
```
leiicamundi: **Blocking (external path) — one audience is shared by two different resource servers.**

On the bundled path the two audiences are correctly distinct: `orchestration-api` here and `camunda-identity-resource-server` at line 120. On the external path both collapse to the single `var.external_oidc.audience`.

That then propagates into the authorization seed, which keys its API declarations and every role permission on `local.oidc.identity.audience` (`identity_authorization.tf:50,63-65,100,107`). With an external provider those permissions end up pointing at the orchestration audience.

Fix: split `external_oidc.audience` into `orchestration_audience` and `identity_audience`.

aws/containers/ecs-single-region-fargate/terraform/cluster/auth_mode.tf
```diff
@@ -0,0 +1,123 @@
...
+    audience                        = string
+    identity_client_id              = string
+    identity_client_secret_arn      = string
+    orchestration_client_id         = string
+    orchestration_client_secret_arn = string
+    connectors_client_id            = string
+    connectors_client_secret_arn    = string
+  })
+  default     = null
+  description = "External OIDC provider config. Optional and only honored when authentication_mode = \"oidc\": when set, that provider is used and the bundled Keycloak is skipped; when null (default), a bundled Keycloak is deployed as the OIDC provider. One client per component (identity, orchestration, connectors); client secrets are existing Secrets Manager ARNs (created out-of-band), never raw values."
```
leiicamundi: **Blocking (external path) — the schema is missing the fields Entra and Okta need.**

The object carries issuer, token URI, audience and one client id + secret ARN per component. Missing, and not settable anywhere else:

- `username_claim` — hardcoded `preferred_username` (`camunda.tf:82`).
- `client_id_claim` — hardcoded `client_id` (`camunda.tf:87`). Entra does not emit `client_id`; it uses `azp` / `appid`.
- Connectors token `scope` — the reference chart renders `camunda.client.auth.scope` from `connectors.authTokenScope`; Entra v2 client-credentials requires `<resource>/.default`, so without it no token is issued.
- The admin principal — hardcoded `"admin"` (`camunda.tf:91` and `identity_authorization.tf:41`). The comment at `camunda.tf:89-90` tells external users to change it, but there is no knob to change.

Either add these four inputs, or narrow the description to say the schema only supports providers that match these fixed claim names. Right now `external_oidc` reads as provider-agnostic and it is not.

(Flagging this as blocking on the external path specifically because the PR describes it as "structurally complete", which reads as ready to use.)

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
```diff
@@ -107,10 +72,47 @@ module "orchestration_cluster" {
...
+      { name = "CAMUNDA_SECURITY_AUTHENTICATION_OIDC_USERNAMECLAIM", value = "preferred_username" },
+      # Detect m2m (client-credentials) callers by the client_id claim; without this,
+      # a service-account token is treated as a user (preferred_username =
+      # service-account-<client>) and never matches the admin/connectors client
+      # mappings below, so deployments are rejected 403. The realm emits client_id.
+      { name = "CAMUNDA_SECURITY_AUTHENTICATION_OIDC_CLIENTIDCLAIM", value = "client_id" },
+      { name = "CAMUNDA_SECURITY_AUTHENTICATION_OIDC_AUDIENCE", value = local.oidc.audience },
+      # Admin user identifier from the username claim (the bundled realm's 'admin';
+      # for external OIDC set this to your admin's preferred_username).
+      { name = "CAMUNDA_SECURITY_INITIALIZATION_DEFAULTROLES_ADMIN_USERS_0", value = "admin" },
```
leiicamundi: The admin principal is hardcoded to `"admin"` here and again as `identity_admin_claim_value` in `identity_authorization.tf:41`. Fine for the bundled realm, which creates exactly that user. Not fine for the external path, where it silently grants admin to whoever happens to have `preferred_username == "admin"` in the customer's directory, or to nobody.

One `admin_claim_value` variable, consumed here and by the Identity mapping rule, keeps the two in sync and makes the external path usable. The two values already have to match, which is a good argument for them being one input rather than two constants.

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
```diff
@@ -181,18 +183,26 @@ module "connectors" {
...
-    }
-  ]
+    ],
+    # Auth to the orchestration cluster: basic user or OIDC client-credentials.
+    # Connectors fetches tokens via the shared ALB (same host as every other actor)
+    # so the token `iss` is the ALB URL and matches what orchestration validates.
+    local.oidc_enabled ? [
+      { name = "CAMUNDA_CLIENT_AUTH_CLIENTID", value = local.oidc.connectors.client_id },
+      { name = "CAMUNDA_CLIENT_AUTH_TOKENURL", value = local.oidc.token_uri },
+      { name = "CAMUNDA_CLIENT_AUTH_AUDIENCE", value = local.oidc.audience },
```
leiicamundi: Two notes on the Connectors OIDC branch.

**Not a bug:** omitting `CAMUNDA_CLIENT_AUTH_METHOD` matches the reference chart, which only sets `method` on the basic branch and relies on auto-detection for OIDC. I checked `templates/connectors/files/_application.yaml`. Setting it explicitly would be nice for symmetry with line 195, but it is not a defect today.

**Gap:** there is no `CAMUNDA_CLIENT_AUTH_SCOPE`. The chart renders `scope` when `connectors.authTokenScope` is set, and Entra v2 rejects a client-credentials request without `<resource>/.default`. Needs an external input before the external path can work.

**Design note:** `local.oidc.token_uri` is the public ALB, so Connectors leaves the VPC through NAT and comes back through the internet-facing load balancer for every token. Deliberate, so that `iss` matches, and it works with the default `limit_access_to_cidrs = ["0.0.0.0/0"]`. It stops working the moment someone narrows that list without whitelisting the NAT egress IP. Worth a line in the README.

aws/containers/ecs-single-region-fargate/terraform/cluster/iam.tf
```diff
@@ -160,12 +160,36 @@ resource "aws_iam_role_policy_attachment" "task_execution_registry" {
...
+      aws_secretsmanager_secret.keycloak_db_password[0].arn,
+      aws_secretsmanager_secret.keycloak_admin_password[0].arn,
+      aws_secretsmanager_secret.realm_admin_user_password[0].arn,
+      aws_secretsmanager_secret.identity_client_secret[0].arn,
+      aws_secretsmanager_secret.orchestration_oidc_client_secret[0].arn,
+      aws_secretsmanager_secret.connectors_oidc_client_secret[0].arn,
+      aws_secretsmanager_secret.keycloak_realm_import[0].arn,
+    ] : [],
+    # External OIDC client secrets (external provider) - customer-provided ARNs.
+    local.use_external ? compact([
```
leiicamundi: External client secret ARNs are added to `ecs_task_secret_arns`, which correctly widens the `secretsmanager:GetSecretValue` resource list. But the `kms:Decrypt` statement further down in the same policy is scoped to `local.secrets_kms_key_arn_effective`, this stack's own CMK.

A customer-supplied secret encrypted with their own CMK will pass the `GetSecretValue` check and fail on decrypt, and it surfaces as a `ResourceInitializationError` at task start rather than anything that points at IAM.

Fix: take the external secrets' KMS key ARNs as inputs and include them in the decrypt statement, or state in the `external_oidc` description that the secrets must use the AWS-managed Secrets Manager key.

Small aside: `compact()` here is redundant, the precondition in `auth_mode.tf:56-71` already rejects empty fields.

aws/containers/ecs-single-region-fargate/terraform/cluster/keycloak_realm.tf
```diff
@@ -0,0 +1,160 @@
...
+    accessTokenLifespan   = 300
+
+    clients = [
+      {
+        clientId                  = "orchestration"
+        name                      = "Orchestration"
+        enabled                   = true
+        protocol                  = "openid-connect"
+        publicClient              = false
+        secret                    = try(random_password.orchestration_oidc_client_secret[0].result, "")
```
leiicamundi: **Should fix — these credentials cannot be changed after the first boot, and nothing says so.**

The three client secrets (lines 87, 103, 115) and the realm admin password (line 137) are embedded in the realm import JSON. Keycloak's startup import skips a realm that already exists, which the module documents at `keycloak/variables.tf:175`.

The consequence is not documented at this level. Replace any of these `random_password` resources, or change a realm URL, and Terraform updates the Secrets Manager value and the consuming task definitions while Keycloak keeps the old realm. Components get the new secret, Keycloak still expects the old one, auth breaks, and the apply exits 0.

Same trap for realm settings: switching `sslRequired` at line 75 by enabling TLS on an existing deployment has no effect on a realm that already exists.

I would not reach for `lifecycle.ignore_changes`, that hides the divergence rather than fixing it. A short "realm bootstrap is create-only, rotate via the admin API then force a new deployment" section in the README is enough for a reference architecture.

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
```diff
@@ -205,3 +215,109 @@ module "connectors" {
...
+  aws_region                  = data.aws_region.current.region
+  s2s_cloudmap_namespace      = module.orchestration_cluster.s2s_cloudmap_namespace
+  log_group_name              = module.orchestration_cluster.log_group_name
+  ecs_task_execution_role_arn = aws_iam_role.ecs_task_execution.arn
+  registry_credentials_arn    = join("", aws_secretsmanager_secret.registry_credentials[*].arn)
+
+  # ALB exposure is opt-in. Flip to true (and confirm the context path) once
+  # Identity should be reachable through the shared ALB.
+  alb_listener_http_webapp_arn         = local.webapp_listener_arn
+  enable_alb_http_webapp_listener_rule = false
```
leiicamundi: `/identity*` is never routed on the ALB, but two places advertise it as if it were:

- `CAMUNDA_IDENTITY_BASE_URL` at line 297 → `http://<alb>/identity`
- the `camunda-identity` client `redirectUris` at `keycloak_realm.tf:120` → `http://<alb>/identity/auth/login-callback`

Right now `/identity*` falls through to the orchestration catch-all at priority 100, so anything following either URL lands on the wrong app.

It is also not just a matter of flipping this flag: the rule forwards `/identity/...` unchanged, and the task sets no matching servlet context path, so the app would receive a path it does not serve.

Either point `CAMUNDA_IDENTITY_BASE_URL` at the Service Connect address (`http://identity:8084`), matching "reachable only in-VPC" in the PR description and dropping the unusable browser callback, or enable the rule together with a context path and test the callback. As written the config describes a topology that does not exist.

aws/containers/ecs-single-region-fargate/terraform/cluster/keycloak.tf
```diff
@@ -0,0 +1,81 @@
...
+  count = local.deploy_bundled_keycloak ? 1 : 0
+
+  depends_on = [null_resource.run_db_seed_task]
+
+  prefix                      = "${var.prefix}-oc1"
+  ecs_cluster_id              = aws_ecs_cluster.ecs.id
+  vpc_id                      = module.vpc.vpc_id
+  vpc_private_subnets         = module.vpc.private_subnets
+  aws_region                  = data.aws_region.current.region
+  s2s_cloudmap_namespace      = module.orchestration_cluster.s2s_cloudmap_namespace
```
leiicamundi: The Service Connect namespace comes from `module.orchestration_cluster`, so Terraform must finish creating the orchestration service before it can start creating Keycloak. Orchestration also runs with `wait_for_steady_state = true` (module default, not overridden). In `oidc` mode orchestration is the component that needs Keycloak's discovery endpoint at boot, so the dependency graph is the reverse of the startup requirement.

It passes CI today because orchestration reaches ECS steady state without a working IdP and retries. That is timing, not ordering.

Moving the shared `aws_service_discovery_http_namespace` up to the cluster root would let both Keycloak and orchestration hang off it, and let Keycloak come up first. That also removes the odd situation where the orchestration module owns a resource three other modules depend on.

aws/containers/ecs-single-region-fargate/terraform/cluster/identity_authorization.tf
```diff
@@ -0,0 +1,145 @@
...
+          { audience = local.oidc.identity.audience, definition = "read" },
+          { audience = local.oidc.identity.audience, definition = "write" },
+        ]
+      },
+    ]
+  }
+
+  # Web Modeler / Camunda Hub. `applications` entries are intentionally omitted: in the
+  # generic OIDC profile the IdP owns the clients (the bundled realm import creates the
+  # `web-modeler` client), and Identity is only a resource server here.
```
leiicamundi: The comment says "the bundled realm import creates the `web-modeler` client". It does not. `keycloak_realm.tf:80-124` defines exactly three clients: `orchestration`, `connectors`, `camunda-identity`.

So enabling this flag declares the Web Modeler resource servers and roles, but nothing in this stack provisions a client that can obtain a token for them. Worth being explicit that the Hub / Web Modeler client is provisioned out of band, otherwise the flag reads as if it makes Hub self-contained.

Dropping `applications` from the preset is the right call and the reasoning in the rest of the comment is sound. It is only the parenthetical that is wrong.

aws/containers/ecs-single-region-fargate/terraform/cluster/identity_authorization.tf
```diff
@@ -0,0 +1,145 @@
...
+# Management Identity ships no roles out of the box — they are declared by
+# `identity.component-presets`. In the generic OIDC profile Identity does not manage
+# the IdP, so the Keycloak-only path for granting them (`keycloak.users[].roles`) is
+# unavailable; roles are granted by matching a token claim through
+# `identity.mapping-rules`. Both keys are honored in the `oidc` profile — the upstream
+# Helm chart renders them outside its `authIssuerType == "KEYCLOAK"` guard.
+#
+# The preset and role definitions below are copied from the reference Helm chart
+# (camunda-platform-helm, charts/camunda-platform-8.10/templates/identity/configmap.yaml)
+# so this reference architecture stays in lockstep with it.
```
leiicamundi: I diffed all 145 lines of this against `charts/camunda-platform-8.10/templates/identity/configmap.yaml` and they match: the `identity` preset (chart lines 185-203), the `webmodeler` apis and the `Web Modeler` / `Web Modeler Admin` roles (chart lines 267-310), and the mapping-rule field names. The claim that both keys render outside the `authIssuerType == "KEYCLOAK"` guard is also correct. Good work, this was the part I most expected to find wrong.

The problem is keeping it that way. "stays in lockstep with it" is a promise with no mechanism: no pinned chart version in the comment, and no assertion over the rendered document. The golden cannot help, `container_definitions` is redacted, as the tfvars comment already notes.

Two cheap anchors: cite the exact chart version you copied from, and add a `terraform test` run asserting a couple of stable strings in `local.identity_authorization_json`. Then a Renovate bump of the Identity image that outgrows these presets has something to trip over.

aws/containers/ecs-single-region-fargate/terraform/cluster/test/golden/golden.tfvars
```diff
@@ -1,2 +1,15 @@
-# use this file for vars without default values
-# for the golden file generation
+# Variables for golden-plan generation.
+#
+# Exercise OIDC mode with the bundled Keycloak (external_oidc unset) so the golden
+# covers the full auth surface this reference adds: the Keycloak service + realm
+# import, Management Identity, and all OIDC wiring. The `basic` default deploys none
+# of that, so testing `oidc` here gives meaningful regression coverage.
+authentication_mode = "oidc"
```
leiicamundi: Pinning the golden to `oidc` is a reasonable trade, it covers far more surface. Worth naming the cost though: the `basic` branch of every `concat()` in `camunda.tf` is now the default users get and the one the golden no longer pins.

The live matrix does deploy both lanes, so this is not a coverage hole. But golden diffs are the cheap, per-PR signal and live runs are the slow one, so a regression in the default path now waits for the expensive check.

If a second golden is too much, `terraform test` at the cluster root would cover it: assert the `basic` env var set, and assert both preconditions in `auth_mode.tf` fire and pass as expected. That also covers the external path, which nothing exercises today.

leiicamundi (commented):
## Review

I read the full diff, checked the Camunda and Keycloak claims against upstream sources, and ran a second pass over the auth wiring.

**Verdict: changes requested.** `basic` and bundled-Keycloak `oidc` look solid and have a live CI lane behind them. The two paths the PR advertises but never ran, TLS and external OIDC, do not work as written.

### Confirmed correct

I checked the claims that would be expensive to get wrong, and they hold:

- `identity.component-presets` and `identity.mapping-rules` really are rendered outside the `authIssuerType == "KEYCLOAK"` guard in `camunda-platform-8.10/templates/identity/configmap.yaml`. Preset bodies, role names and the `claim-name` / `claim-value` / `operator` / `rule-type` / `applied-role-names` schema match the chart line for line, and the four audiences are the chart defaults.
- Not setting `CAMUNDA_CLIENT_AUTH_METHOD` in OIDC mode matches the chart, which only sets `method` on the basic branch. I went looking for a bug here and there isn't one.
- Probing `/auth/health/ready` on port 9000 is right: the Keycloak management interface inherits `http-relative-path`.
- The golden plan is clean. No account IDs, ARNs, emails or secret values.

### Blocking

1. **TLS cannot work.** `alb_base_url` is hardcoded to `http://` and never consults `alb_https_enabled`, so setting `alb_certificate_arn` leaves the issuer, the redirect URI and the realm root on HTTP while every request gets 301'd to HTTPS. Keycloak then advertises an `https://` issuer and Camunda rejects it at startup. Separately there is no input for the certificate's hostname, and ACM will not issue a public certificate for an `*.elb.amazonaws.com` name. This is the path offered as the answer to the earlier Major finding, so it needs to actually run.

2. **External OIDC is missing the inputs a real provider needs.** One shared `audience` for both Orchestration and Identity, no `username_claim`, no `client_id_claim`, no Connectors `scope`, no configurable admin principal, and `kms:Decrypt` scoped to the local CMK only. Against Entra or Okta several of those fail at once. Either add the inputs, or say in the variable description that the schema only targets providers matching these fixed assumptions.

### Should fix

3. **Realm content never converges.** `--import-realm` skips a realm that already exists, and the client secrets plus the realm admin password live inside that JSON. Replacing any of them updates Terraform and Secrets Manager but not Keycloak, and the apply still reports success.

4. **`/identity*` is advertised but not routed.** `enable_alb_http_webapp_listener_rule` is `false`, yet `CAMUNDA_IDENTITY_BASE_URL` and the `camunda-identity` client callback both point at `http://<alb>/identity`.

5. **Create ordering is inverted.** Keycloak consumes the Service Connect namespace the orchestration module creates, and orchestration waits for steady state, so Keycloak is created after the component that needs it at boot. It works today, but only because of startup timing.

6. **The preset copy has no anchor.** 145 lines copied from the chart, with no pinned chart version and nothing asserting the rendered document, so an Identity image bump can leave it stale without any signal.

### Nits

7. The comment at `identity_authorization.tf:71` says the realm import creates a `web-modeler` client. It creates `orchestration`, `connectors` and `camunda-identity`.
8. Pinning the golden to `oidc` drops the default `basic` plan shape from golden coverage. The live matrix still covers it, so this is only a note.

### Review loop

CI is green on both lanes with no skip labels, and the seven Copilot findings were all fixed and answered on thread. But every one of those threads is now `outdated`: the branch was rewritten after that review, and around thirty commits, including the whole external-OIDC, TLS and authorization-seed work, have never been looked at by it. Worth re-requesting Copilot before this goes to a human. `hisImminence`'s changes-requested is also still the standing decision.

Detail is in the inline comments.

hisImminence (approved):
Reviewed the full diff. Solid PR — Secrets Manager/CMK usage, ALB listener-rule priorities, SG scoping, `sslRequired` gating, SQL-identifier validation, IAM scoping, and golden-plan redaction all check out.

One note (not blocking): `terraform/cluster/camunda.tf` deploys `management_identity` with `enable_alb_http_webapp_listener_rule = false` by default, but the bundled realm's `camunda-identity` Keycloak client (`keycloak_realm.tf`) is unconditionally configured with `standardFlowEnabled = true` and a `redirectUris` pointing at `${identity_public_base}/auth/login-callback` — a path nothing routes today. This is already called out in the PR description's "Known limitations" section and nothing in this PR's scope exercises that browser flow, so it's inert rather than broken. Worth carrying the same callout into the committed README (`aws/containers/ecs-single-region-fargate/README.md`, "Authentication modes" section) so it isn't missed when the stacked Web Modeler/Hub work flips the ALB rule on.

aws/containers/ecs-single-region-fargate/terraform/cluster/auth_mode.tf
marcel-dias: Fixed in cb44fd19 — `external_oidc.audience` is split into `orchestration_audience` and `identity_audience`.

You were right about the propagation path: `local.oidc.identity.audience` now resolves from `identity_audience`, so the resource-server declaration at `identity_authorization.tf:50` and the role permissions at 63-65, 100 and 107 point at Identity's own API rather than at the orchestration one. The required-field precondition and its error message were updated to match.

aws/containers/ecs-single-region-fargate/terraform/cluster/auth_mode.tf
marcel-dias: Fixed in cb44fd19. The object now carries each of the four:

- `username_claim` and `client_id_claim`, both `optional()` defaulting to the previous hardcoded `preferred_username` / `client_id`, resolved through `local.oidc` and consumed at `camunda.tf:82,87`. Your Entra point is the reason the second one exists — a provider that emits `azp`/`appid` would otherwise have every m2m token treated as a user and rejected 403.
- `connectors_token_scope`, emitted as `CAMUNDA_CLIENT_AUTH_SCOPE` only when non-empty, so Keycloak keeps issuing tokens without a scope and Entra v2 can be given `<resource>/.default`.
- The admin principal, lifted to a top-level `admin_claim_value` rather than into this object, because the bundled path needs it too. See the reply on the `camunda.tf` thread.

The client-secret KMS constraint you raised separately is now stated in this variable's description as well (8aad3d96).

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
marcel-dias: Fixed in cb44fd19 — one `admin_claim_value` variable, defaulting to `"admin"`.

It is consumed in both places you identified: `CAMUNDA_SECURITY_INITIALIZATION_DEFAULTROLES_ADMIN_USERS_0` here, and `local.identity_admin_claim_value` in `identity_authorization.tf`, which feeds the mapping rule and `IDENTITY_INITIAL_CLAIM_VALUE`. The claim *name* was given the same treatment and now resolves from `local.oidc.username_claim`, so a provider whose claim is not `preferred_username` works without editing either file.

Your argument that two values which must match should be one input is what drove the shape of this.

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
marcel-dias: Thanks for checking `_application.yaml` — I took the "not a bug" half at face value and left `CAMUNDA_CLIENT_AUTH_METHOD` unset on the OIDC branch, matching the chart's reliance on auto-detection.

The gap half is fixed in cb44fd19: `CAMUNDA_CLIENT_AUTH_SCOPE` is rendered from `local.oidc.connectors.token_scope`, which is empty for the bundled realm (so no env var is emitted and nothing changes there) and settable via `external_oidc.connectors_token_scope` for a provider that requires one.

On the design note about `token_uri` being the public ALB: left as is deliberately, so the token `iss` matches what orchestration validates. Happy to revisit if you would rather split the issuer from the token endpoint.

aws/containers/ecs-single-region-fargate/terraform/cluster/iam.tf
marcel-dias: Taken the documentation option you offered rather than the input one, in 8aad3d96 — and the first wording I pushed was wrong, so this is worth reading.

I initially wrote that the customer CMK should be "added to `var.secrets_kms_key_arn`". That input is a single key that *replaces* the CMK this stack would otherwise create and encrypt its own secrets with, so following that advice would re-key the whole stack rather than widen the grant. The description now states the real constraint: `kms:Decrypt` is scoped to one key, so the external client secrets must be encrypted either with the AWS-managed Secrets Manager key or with that same key, and anything else is readable but not decryptable — surfacing exactly as the `ResourceInitializationError` you described.

The multi-key input remains the better fix and is not implemented; this is the documented constraint only.

aws/containers/ecs-single-region-fargate/terraform/cluster/keycloak_realm.tf
marcel-dias: Not addressed — agreed it should be, and I have not done it.

Confirmed the behaviour: `--import-realm` skips a realm that already exists, so replacing any of the three client-secret `random_password` resources or the realm admin password updates Secrets Manager and the consuming task definitions while Keycloak keeps serving the old realm. Components get the new secret, Keycloak still expects the old one, and the apply exits 0. The module-level note at `keycloak/variables.tf:175` does not carry that consequence up to where the secrets are generated.

Leaving it open rather than adding a comment I would then have to defend as sufficient: a warning does not stop the silent breakage, and the real fix is either a realm-update path or making the rotation an explicit, documented destroy/recreate.

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
marcel-dias: Not addressed.

Verified it is exactly as you describe: `enable_alb_http_webapp_listener_rule = false` for Management Identity at line 253, while `CAMUNDA_IDENTITY_BASE_URL` at line 307 and the `camunda-identity` client's `rootUrl` / `redirectUris` at `keycloak_realm.tf:119-120` all still resolve from `local.identity_public_base` (`<alb>/identity`). Anything following either URL lands on the orchestration catch-all.

Your second point is the reason I have not just flipped the flag: the rule would forward `/identity/...` unchanged and the task serves no matching context path, so enabling it alone moves the failure rather than fixing it. Pointing `CAMUNDA_IDENTITY_BASE_URL` at the Service Connect address is the smaller, coherent change, but it needs the realm redirect URI reconciled with it, so I would rather do it deliberately than fold it into this round.

aws/containers/ecs-single-region-fargate/terraform/cluster/keycloak.tf
marcel-dias: Not addressed.

Confirmed: `keycloak.tf:21` takes `s2s_cloudmap_namespace` from `module.orchestration_cluster`, so Terraform must create the orchestration service before it starts Keycloak, while in `oidc` mode orchestration is the component that wants Keycloak's discovery endpoint at boot. And you are right that CI passing proves timing rather than ordering — orchestration reaches steady state without a working IdP and retries.

Hoisting the shared `aws_service_discovery_http_namespace` to the cluster root is the right shape and I have not done it: it moves a resource that both modules currently depend on, which is a state migration on existing deployments, and I did not want to bundle that with the auth fixes in this round.

aws/containers/ecs-single-region-fargate/terraform/cluster/identity_authorization.tf
marcel-dias: Fixed in cb44fd19 — thank you for catching it, the parenthetical was plainly wrong.

`keycloak_realm.tf` defines exactly `orchestration`, `connectors` and `camunda-identity`; there is no `web-modeler` client anywhere in the import. The comment now says so and states that a client able to obtain a token for these audiences has to be provisioned out of band, so the flag no longer reads as if it makes Web Modeler self-contained.

aws/containers/ecs-single-region-fargate/terraform/cluster/identity_authorization.tf
marcel-dias: Thank you for diffing all 145 lines — that is the check I most wanted someone to run independently.

The lockstep concern is not addressed, and your reasoning for why the usual safety nets do not help is correct: the golden cannot pin it, because the AWS provider marks `container_definitions` sensitive and the rendered `SPRING_APPLICATION_JSON` is redacted out of the plan entirely.

I have not added a mechanism because the honest options are all heavier than a comment: pinning a chart version in prose still drifts silently, and the only check with teeth is an assertion over the rendered document — which means rendering it outside the task definition so something other than ECS can see it. Worth doing, but as its own change rather than tacked onto this one.

aws/containers/ecs-single-region-fargate/terraform/cluster/test/golden/golden.tfvars
marcel-dias: Not changed, and the cost you name is real: with the golden pinned to `oidc`, the `basic` branch of every `concat()` in `camunda.tf` is both the default users get and the one the cheap per-PR signal no longer covers.

Your `terraform test` suggestion is the right answer — it would cover the `basic` env set and both `auth_mode.tf` preconditions without a second golden, and it is the only thing that would exercise the external path at all, which nothing currently does. I have not written it in this round.

One update since you commented: the preconditions have grown, so such a test now has more to assert than when you raised this — the TLS certificate/hostname pair and the `enable_camunda_hub` + `basic` combination both fail fast now too.

aws/containers/ecs-single-region-fargate/terraform/cluster/variables.tf
```diff
@@ -101,6 +113,65 @@ variable "db_seed_iam_usernames" {
   default     = ["camunda"]
 }
```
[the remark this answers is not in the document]
marcel-dias: Fixed in 3603fa5f.

Correct that this one was missed: `db_name`, `identity_db_name`, `identity_db_username`, `keycloak_db_name` and `keycloak_db_username` all received the PostgreSQL-identifier guard in an earlier round, and `db_seed_iam_usernames` did not — even though it is interpolated the same two ways you describe, and additionally drives the `IAM_DB_USERS` shell loop.

Every element is now validated against the same rule (`^[a-zA-Z_][a-zA-Z0-9_]*$`, ≤ 63 characters) via `alltrue`, so a quote or whitespace in any entry fails at plan time rather than reaching the seed task.

aws/containers/ecs-single-region-fargate/terraform/cluster/variables.tf
```diff
@@ -55,6 +55,10 @@ variable "ports" {
     zeebe_gateway_cluster_port            = 26502
     zeebe_gateway_network_port            = 26500
     zeebe_broker_network_command_api_port = 26501
+    management_identity_app               = 8084
+    management_identity_management        = 8082
+    keycloak_http                         = 18080
+    keycloak_management                   = 9000
```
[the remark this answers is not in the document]
marcel-dias: Fixed in 3603fa5f.

The security-group port map is now composed conditionally, exactly along the lines you describe: the two `management_identity_*` entries are kept only when `local.oidc_enabled`, and `keycloak_http` / `keycloak_management` only when `local.deploy_bundled_keycloak`. So basic mode opens neither, and an external-provider OIDC deployment opens Identity's ports but not Keycloak's.

One deliberate non-result worth recording: the golden plan is unchanged. That is expected rather than an oversight — it runs with `oidc` and the bundled Keycloak, so all four ports legitimately stay open there and the security group is identical. The paths this narrows are `basic` and external-provider, neither of which the golden covers.

aws/containers/ecs-single-region-fargate/terraform/cluster/keycloak.tf
```diff
@@ -0,0 +1,81 @@
...
+  secrets = [
+    { name = "KC_DB_PASSWORD", valueFrom = aws_secretsmanager_secret.keycloak_db_password[0].arn },
+    { name = "KC_BOOTSTRAP_ADMIN_PASSWORD", valueFrom = aws_secretsmanager_secret.keycloak_admin_password[0].arn },
+    # The camunda-platform realm (clients + admin user, with their secrets) is
+    # imported on startup from this JSON — Keycloak self-provisions the realm.
+    { name = "KEYCLOAK_REALM_IMPORT_JSON", valueFrom = aws_secretsmanager_secret.keycloak_realm_import[0].arn },
+  ]
+
+  # Write KEYCLOAK_REALM_IMPORT_JSON to the import dir and start with --import-realm.
+  enable_realm_import = true
```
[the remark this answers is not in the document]
marcel-dias: Not addressed, and you are right about the mechanism.

This is the same finding `leiicamundi` raised on `keycloak_realm.tf:87`, where I replied in more detail: `--import-realm` skips a realm that already exists, so rotating any of the three client secrets or the realm admin password updates Secrets Manager and the consuming task definitions while Keycloak keeps serving the old realm — components get the new value, Keycloak still expects the old one, and the apply exits 0.

Leaving it open rather than adding a comment I would then have to defend as sufficient: a warning does not prevent the silent breakage. The real fix is either a realm-update path or making rotation an explicit, documented destroy/recreate, and that belongs in its own change.

aws/containers/ecs-single-region-fargate/terraform/cluster/auth_mode.tf
```diff
@@ -0,0 +1,189 @@
...
+          var.external_oidc.orchestration_audience,
+          var.external_oidc.identity_audience,
+          var.external_oidc.identity_client_id,
+          var.external_oidc.identity_client_secret_arn,
+          var.external_oidc.orchestration_client_id,
+          var.external_oidc.orchestration_client_secret_arn,
+          var.external_oidc.connectors_client_id,
+          var.external_oidc.connectors_client_secret_arn,
+        ] : trimspace(v) != ""
+      ])
```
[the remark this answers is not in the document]
marcel-dias: Checked this one rather than taking it at face value, and it does not reproduce — Terraform short-circuits `||` here, so the default path is fine.

I built an isolated configuration with the same shape (an `object` variable defaulting to `null`, a `terraform_data` precondition whose condition is `var.x == null || alltrue([for v in [var.x.attr, ...] : trimspace(v) != ""])`) and planned it with the variable left at `null`: `Plan: 1 to add`, no error. The attribute accesses on the right-hand side are not evaluated when the left-hand side is true.

The existing golden plan is corroborating evidence: it runs with `authentication_mode = "oidc"` and `external_oidc` unset — that is, null — and plans 159 resources successfully through this exact precondition on every regeneration.

Leaving the condition as written. Happy to reconsider if you have a Terraform version where the eager-evaluation behaviour does reproduce.

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
marcel-dias: Now fixed in the commit above — I had replied that this was not addressed, and you asked for it, so here is what changed.

Both of your points held up when I traced them:

- Nothing routed `/identity*`. The module gates *both* the target group and the listener rule on `enable_alb_http_webapp_listener_rule`, which was `false`, so no rule existed and `/identity/...` fell through to the orchestration catch-all at priority 100.
- Enabling the flag alone would not have fixed it. ALB's `forward` action does not rewrite the path, the task had no matching prefix, and it would have answered 404 — which is exactly what the variable description meant by "requires a decided context path".

So the path is now decided and applied in all three places that have to agree: `context_path` is a module input (default `/identity`) that the listener rule matches as `"<context_path>*"`; the task receives `SERVER_SERVLET_CONTEXT_PATH`, which is the key the reference Helm chart renders from `identity.contextPath`; and `local.identity_context_path` is the single source of truth that `identity_public_base` — and therefore `CAMUNDA_IDENTITY_BASE_URL` and the realm `rootUrl`/`redirectUris` — is built from. The routed path and the advertised URL can no longer drift.

Health checks are deliberately untouched: both the target group probe and the container probe use the management port (8082, `/actuator/health/liveness`), which does not carry the servlet context path. The chart documents the same asymmetry, noting Identity does not support a context path for health endpoints.

On the ordering itself: it was already correct and is now actually exercised — identity at 30 ahead of the catch-all at 100, with keycloak (40) and connectors (50) on disjoint patterns. Confirmed in the regenerated golden, which gains exactly the target group and the listener rule and moves no other resource.

aws/containers/ecs-single-region-fargate/terraform/cluster/camunda.tf
```diff
@@ -205,3 +225,109 @@ module "connectors" {
...
+    # First admin is granted by matching this claim/value (write-once at first boot).
+    { name = "IDENTITY_INITIAL_CLAIM_NAME", value = local.identity_admin_claim_name },
+    { name = "IDENTITY_INITIAL_CLAIM_VALUE", value = local.identity_admin_claim_value },
+    ],
+    # Identity's own authorization model (roles + claim-based grants). Opt-in, because
+    # it only matters once a component that resolves permissions through Identity is
+    # deployed; see identity_authorization.tf.
+    local.webmodeler_authorization_enabled ? [
+      { name = "SPRING_APPLICATION_JSON", value = local.identity_authorization_json },
+    ] : [],
```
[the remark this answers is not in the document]
marcel-dias: Fixed in d7d7450c — good catch, and a consequential one: this failure mode is silent.

Verified the setup is exactly as you describe. `IDENTITY_INITIAL_CLAIM_NAME`/`VALUE` and the seeded `Camunda Admin` rule are both built from `local.identity_admin_claim_name` and `local.identity_admin_claim_value`, and the seeded rule is `EQUALS`/`ROLE` — so the auto-created `Default` rule carries an identical triple. The task stays healthy either way, so nothing surfaces until a Web Modeler request is denied.

The bootstrap pair is now emitted only when the authorization seed is off.

Two things that make this safe rather than a trade:

- Nothing is lost when the seed is on. `identity_admin_role_names` is the union of the role names from both presets, so the seeded rule grants `ManagementIdentity` as well as `Web Modeler` and `Web Modeler Admin` — a strict superset of what the bootstrap claim grants. The fix holds even where the de-duplication details differ from the description.
- It matches the reference implementation: the Helm chart never sets `IDENTITY_INITIAL_CLAIM_*` at all and grants through mapping rules alone. I checked `charts/camunda-platform-8.10/templates/identity/deployment.yaml` for this.

Worth recording that the golden is unchanged by this, and structurally cannot change: the AWS provider marks `container_definitions` sensitive, so the task environment is redacted out of the plan. That blind spot is documented in `golden.tfvars`, and this finding is a good argument for the end-to-end test that would actually cover it.

marcel-dias:
## Camunda Hub browser E2E on a live environment: 2 of 10 scenarios pass

Deployed this branch (`ca914c06`) to a throwaway environment and drove the 10 browser scenarios with Playwright. **The Hub authorization path does not work end to end**, despite `enable_camunda_hub_authorization = true`.

**Environment**: `prefix=ecs-hub-demo`, `eu-central-1`, `authentication_mode=oidc`, `enable_camunda_hub=true`, `enable_camunda_hub_authorization=true`. `terraform apply` clean: **183 added, 0 changed, 0 destroyed**. All five ECS services reached their desired count (orchestration 3/3; keycloak, management-identity, camunda-hub, connectors 1/1). Environment destroyed after the run.

### Results

| # | Step | Result |
|---|---|---|
| 1 | `GET /hub` -> Keycloak authorize (PKCE) | ✅ redirect with `client_id=web-modeler`, `code_challenge_method=S256`, callback `/hub/login-callback` |
| 2 | Sign in as realm user `admin` | ✅ first attempt; `iss` matches the ALB URL |
| 3 | Modeler Home | ❌ two error toasts, 10 console errors |
| 4 | Projects tab | ❌ empty state renders, but behind an error toast rather than clean |
| 5 | Create project | ❌ `POST /hub/api/internal/projects` -> **404** |
| 6 | Collaborators | ⛔ blocked by 5 |
| 7 | Create BPMN diagram | ⛔ blocked by 5 |
| 8 | Reload Home, project persists | ⛔ blocked by 5 |
| 9 | API probes | ❌ see below |
| 10 | `/hub/clusters/default-cluster` | ❌ bounces to `/hub/login?returnUrl=...`, back to Home with the same toasts |

Toasts: *"Yikes! Could not fetch your projects."* and *"Yikes! Could not fetch your shared resources."*

### API evidence

```
GET  /hub/api/internal/login                                          200
GET  /hub/api/internal/self                                           200
GET  /hub/api/internal/clusters                                       200
GET  /hub/api/external/identity/permissions/for-token
       ?audience=web-modeler-api                                      404   <-- key
GET  /hub/api/internal/management/clusters                            403
GET  /hub/api/internal/organizations/<org>/projects?includeDetails    404
POST /hub/api/internal/projects                                       404
POST /hub/api/internal/pusher/channel-authorizations                  401
```

Authentication is fine (`login`, `self`, `clusters` all 200). Everything that needs an authorization decision from Management Identity fails. This is the exact `403`/`404` signature this PR's own description predicts for the flag being *off*, except the flag is on.

### What I ruled out

The Terraform side is correct, which narrows this considerably:

- **The seed document is delivered and correct.** Read back from the running task definition, `SPRING_APPLICATION_JSON` contains both presets (`identity` -> `camunda-identity-resource-server` / `ManagementIdentity`; `webmodeler` -> `web-modeler-api`, `web-modeler-public-api` / `Web Modeler`, `Web Modeler Admin`) and the `Camunda Admin` mapping rule (`preferred_username` = `admin`, `EQUALS`/`ROLE`) granting all three roles.
- **`IDENTITY_INITIAL_CLAIM_NAME`/`_VALUE` are absent from the task**, so the shadowing fix in `d7d7450c` works as intended and is not the cause.
- **The realm audience mappers are wired.** `webmodeler_client.protocolMappers` includes `web-modeler-api-audience` and `web-modeler-public-api-audience`, both `oidc-audience-mapper` with `access.token.claim=true`.
- **Service Connect is enabled** on keycloak, identity and camunda-hub, all in namespace `ns-j7krvvo47zpd6qwp`, with aliases `keycloak:18080` and `identity:8084`.

### Probable root cause: Identity cannot reach its IdP backend

Management Identity logs an unresolvable address on every SDK call:

```
ERROR i.s.e.RestResponseEntityExceptionHandler : Unexpected error
io.camunda.identity.sdk.impl.rest.exception.RestException: request failed
Caused by: java.net.ConnectException
Caused by: java.nio.channels.UnresolvedAddressException
```

Its backend URL is the Service Connect alias:

```
CAMUNDA_IDENTITY_ISSUER_BACKEND_URL = http://keycloak:18080/auth/realms/camunda-platform
```

Probing Identity directly with a valid `camunda-identity` m2m token (`aud: [camunda-identity-resource-server, account]`) returns **500** on `/identity/api/roles`, `/identity/api/apis` and `/identity/api/mapping-rules`, with those same stack traces. So Identity is up and authenticating, but cannot complete the IdP calls needed to resolve permissions, which is consistent with Hub's 403/404s.

I could not confirm *which* hostname fails to resolve, because `task_enable_execute_command` is false on the Identity service, so there is no shell in the task. **Next diagnostic step**: enable ECS Exec for that service and run `getent hosts keycloak` plus a curl to `http://keycloak:18080/auth/realms/camunda-platform/.well-known/openid-configuration` from inside the task. That distinguishes "Service Connect alias not resolvable from this task" from "resolvable but the wrong URL".

### Two smaller observations

- The header link "Camunda Modeler" points at **`/hub/hub/`**, a duplicated context path. Cosmetic, but it suggests a base-path is being applied twice somewhere in the Hub configuration.
- One navigation produced a `302` to **`localhost`** (`ERR_CONNECTION_REFUSED`). I could not reproduce it deliberately, so treat it as unconfirmed, but a `localhost` redirect target would point at an unset public URL somewhere.

### Why CI did not catch this

The OIDC CI lane now deploys Hub and the seed, and it passes. It asserts that services reach steady state, not that an authorization check succeeds. Management Identity stays **healthy** throughout this failure, because its probes hit the management port. The golden cannot see it either, since `container_definitions` is redacted from the plan.

So this is not a regression that slipped past the tests; it is a gap the tests cannot currently observe. A request that creates a project is the smallest assertion that would have caught it.

marcel-dias:
## Camunda Hub browser E2E: 10 of 10 scenarios pass on 8.10.0-alpha5

Follow-up to the earlier run that failed 8 of 10. The cause was found, fixed in `e6c11189`, and the full flow re-verified on a fresh environment.

### Root cause

`d32c7905` exposed the Management Identity UI on the shared ALB, which required giving the task `SERVER_SERVLET_CONTEXT_PATH` so it answers the `/identity*` rule. That moved **every** Identity route under the prefix. Camunda Hub still reached Identity in-VPC at the bare Service Connect address, so its requests arrived at the application but matched no route, and every Identity call returned 404.

The symptom read as a permissions problem rather than a routing one: Hub authenticated fine (`login`, `self`, `clusters` all 200) while `identity/permissions/for-token` returned 404, the management API returned 403 and org-scoped projects returned 404. Both tasks stayed **healthy** the entire time, because their probes use the management port, which the servlet context path does not apply to. Neither the golden nor the integration suite could see it: the golden redacts `container_definitions`, and the suite asserts steady state rather than authorization.

The fix builds the in-VPC URL from the same local that sets the context path on the task and the path pattern on the listener rule, so the three cannot drift apart again.

### Results

Environment: fresh `terraform apply` (183 resources), `eu-central-1`, `authentication_mode=oidc`, `enable_camunda_hub=true`, `enable_camunda_hub_authorization=true`, all five services at desired count (orchestration 3/3).

| # | Step | Result |
|---|---|---|
| 1 | `GET /hub` -> Keycloak authorize (PKCE) | ✅ `client_id=web-modeler`, `code_challenge_method=S256` |
| 2 | Sign in as realm user `admin` | ✅ first attempt, `iss` matches the ALB |
| 3 | Modeler Home | ✅ no error toasts |
| 4 | Projects tab | ✅ clean empty state |
| 5 | Create project | ✅ `de468b51-48f4-4ea0-a642-ab73138b9c26` |
| 6 | Collaborators | ✅ `admin@example.com` = Project Admin |
| 7 | Create resource in the project | ✅ process application `ab907652-8c77-4ef9-845e-830d0c146d07` |
| 8 | Reload Home | ✅ project persists, no empty state |
| 9 | API probes | ✅ all 200 |
| 10 | `/hub/clusters/default-cluster` | ✅ Status **Healthy**, Zeebe **Healthy 8.10.0** |

Every call that previously failed now returns 200:

```
/hub/api/external/identity/permissions/for-token?audience=web-modeler-api   404 -> 200
/hub/api/internal/management/clusters                                       403 -> 200
/hub/api/internal/organizations/<org>/projects?includeDetails=true          404 -> 200
/hub/api/internal/management/clusters/default-cluster/usage-metrics                200
/hub/api/internal/pusher/channel-authorizations                             401 -> 200
```

### Also in this push

- **`752750c6`** moves all five Camunda images from `8.10.0-alpha3` to `8.10.0-alpha5`, verified released for each before bumping. Keycloak is unchanged. `camunda/identity` and `camunda/connectors-bundle` also publish an `alpha5.1` patch tag, deliberately not taken so all five stay on one line.
- **`c034e3e6`** regenerates the golden for AWS provider v6.65.0, which adds `replica` to `aws_secretsmanager_secret`. Same class of drift as the `warning_event_categories` refresh: no lock file is committed for this root module, so the stored plan goes stale whenever a provider release lands.

### Notes for whoever maintains the test scenarios

- On alpha5 an empty project no longer offers **Create BPMN diagram** directly; the entry point is **Create process application**, which is the path step 7 now takes. The dialog pre-selects `default-cluster (8.10.0)`, which is itself a useful signal that cluster discovery works.
- Two earlier observations are resolved rather than outstanding: the `401` on Pusher channel authorizations is gone, and the `UnresolvedAddressException` seen in the Identity log was an artifact of probing non-existent API paths, not a connectivity fault. Service Connect resolution was confirmed from inside the task: the alias is present in `/etc/hosts` and the discovery endpoint responds.
- Still cosmetic and unexplained: the header "Camunda Modeler home" link points at `/hub/hub/`, a duplicated context path.

### Coverage gap this leaves

The integration suite deploys Hub and the seed, and it passed while this bug was live. It asserts that services reach steady state, not that an authorization decision succeeds. The smallest assertion that would have caught this is a request that creates a project.

